Issue #5, opened 2020-03-19
Implement header with drop down navigation menu


Manvel:
# Background
It’s a good idea to split big tasks into smaller ones when possible. This way we can prevent future mistakes, work on a separate component, make the review simpler and faster.

For that reason we can start with the header as part of the https://github.com/browser-automation/cba-website/issues/1 issue as it persists on almost all pages and we can use a lot of techniques we have learned already, like responsive web design.

The header itself consist of the website title, subtitle and navigation bar with subitems, as shown in the image below:

<img width="987" alt="Screen Shot 2020-03-18 at 7 47 41 PM" src="https://user-images.githubusercontent.com/716969/77052341-81db7e80-69ee-11ea-94fa-cc4b14bd0db2.png">

As the first step the items can be hardcoded, but eventually we will populate them during the build dynamically using [front-matter variables](https://cmints.io/en/documentation/pages/frontmatter) in [EJS](https://cmints.io/en/documentation/themes/ejs).

Currently the navigation bar is implemented as a native select box on smaller screens, this is fine as the first step, but eventually we might want to find a better solution - like a slidable sidebar menu attached to the hamburger menu, which we can tackle separately.

# What to change
Implement header from https://chrome-automation.com/ with drop down navigation menu.

Manvel:
We have implemented the minimum requirement for Header menu, for further accessibility improvements we will continue in https://github.com/browser-automation/cba-website/issues/17

PR #7, opened 2020-03-31
Issue 5

naarrek:
My very first version of the header part.


theme/layouts/_navigation.ejs
```diff
@@ -1,9 +1,69 @@
...
+        
+      </ul>
+    </li>
+    <li class="menu-item has-children"><a href="">What's New <span class="dropdown-icon"></span></a>
+      <ul class="sub-menu">
+        <li class="menu-item"><a href="">Version 8.0</a></li>
+        <li class="menu-item"><a href="">Version 6.0</a></li>
+        <li class="menu-item"><a href="">Version 5.0</a></li>   
+      </ul>
+    </li>
```
Manvel: I assume you was planning to, but I'll leave the comment just in case:

You can use `EJS` to simplify the HTML generation instead of hardcoding whole HTML code here. For that you will need a data-structure to generate menu structure from and create EJS code for extracting those (We also need to consider nesting multiple level sub menus).

In EJS you can use functions, different loops to create that.

Here are some links you might find useful:

- EJS examples -> https://github.com/mde/ejs/tree/master/examples
- EJS playground -> https://ionicabizau.github.io/ejs-playground/
- EJS Syntax -> https://github.com/mde/ejs/blob/master/docs/syntax.md

theme/layouts/_navigation.ejs
```diff
@@ -1,9 +1,69 @@
-<nav>
-  <ul>
-    <% for (const {title, link} of site.navigation) {%>
-      <li>
-        <a href="<%- link %>" target="_blank"><%- title %></a>
-      </li>
-    <% } %>
+<nav class="nav nav--cen">
+  <ul class="menu">
+    <li class="menu-item has-children"><a href="#">Home</a></li>
```
Manvel: Suggestion: You don't have to specify classnames for identifying menu items you can do that also by `nav li` considering that all list items in current example are "menu-item", unless you plan to use "menu-item" class somewhere else, which doesn't seem to be the case in this situation.

Manvel: Suggestion: You have specified "has-children" class on all top level list items, you can use `nav > ul > li` instead, that way we will be able to avoid ambiguity as in this case, when the item class set to "has-children" while it doesn't have sub items.

theme/layouts/_navigation.ejs
```diff
@@ -1,9 +1,69 @@
...
+    <li class="menu-item has-children"><a href="">Documenttion <span class="dropdown-icon"></span></a>
+      <ul class="sub-menu">
+        <li class="menu-item"><a href="">Interface<span class="dropdown-icon"></span></a>
+              <ul class="sub-menu">
```
Manvel: Detail/suggestion: Here you can also use something like `nav > ul ul` to select all "UL" items which are not the top level one, that's what current ".sub-menu" class meant for I guess.

theme/less/_navigation.less
```diff
@@ -1,26 +1,265 @@
...
-      a
-      {
-        &:hover
-        {
-          text-decoration: underline;
+// 
+// The Menu-list
+//
+
+.menu {
```
Manvel: Suggestion: As we have only 1 navigation menu, I would rather suggest to use an ID(ex.: "#main-navigation" or "#top-navigation"), or to make it semantically more understandable, you can use a selector `header nav` for example and maybe renaming the file to "_header.less".

theme/less/_navigation.less
```diff
@@ -1,26 +1,265 @@
...
+
+  //Make the menu-item take all available space
+  .menu-item {
```
Manvel: See [comments in _navigation.ejs](https://github.com/browser-automation/cba-website/pull/7/files#diff-18f055c29e5211ab0d148460aa1716e9R3), that will also apply to classes below like "has-children" and etc.

theme/less/_navigation.less
```diff
@@ -1,26 +1,265 @@
...
+    display: block;
+
+// Make the menu-items handle the fly-outs smoothly by not allowing
```
Manvel: Suggestion/Detail: let's align comments with the code piece to make it more readable.

theme/less/_navigation.less
```diff
@@ -1,26 +1,265 @@
...
+//
+
+.menu {
+
+  //Make the menu-item take all available space
+  .menu-item {
+    display: block;
+
+// Make the menu-items handle the fly-outs smoothly by not allowing 
+//    break the flow
```
Manvel: Detail: The comment here(which I assume is from the tutorial you have mentioned) would like to describe relative and absolute positioned items relations. So if we organize the CSS file in more readable way we won't need this comment at all I guess, so we will understand that there is a relative > absolute relationship between items.

theme/less/_variables.less
```diff
@@ -1,4 +1,41 @@
...
+// 2 //
+@menu-co: #ddd;
+@menu-co-hover: #fff;
+
+// 3 //
+@menu-bo: rgba(white, .1);
+
+// 4 //
+@sub-menu-bg: #eee;
+@sub-menu-bg-hover: white;
```
Manvel: Why are we using those colors ? Are you we going to make the menu look like as it's currently on chrome-automation.com eventually, or the plan is to change whole design of the website while we are rewriting it?

theme/less/_navigation.less
```diff
@@ -1,26 +1,265 @@
...
+      content: "+";
+      display: block;
+      transition: transform .25s ease-in-out;
+    }
+  }
+}
+
+// Adjust the direction and dropdown-icon positioning for the 
+// right-to-left menu
+.nav--rtl {
```
Manvel: I'm not sure when we will decide to translate the website to support RTL, not in near future. We can try creating minimal support for RTL, but if we are going to go in that route there is more to this support, for example the direction in which the 2-nd level submenu items are being opened that needs to be switched as well.

theme/less/_navigation.less
```diff
@@ -1,26 +1,265 @@
...
+  }
+  .nav--rtl .menu .sub-menu .sub-menu {
+    left: auto;
+    right: 100%;
+  }
+
+  .nav--cen .menu {
+    justify-content: center;
+  }
+}
```
Manvel: Narek jan you did a great job for starting this PR and I know you have put a lot of effort into it, it's great start, but I think we might want to change our approach to the current issue and try creating something that is relevant for current project and expand to make it generic as soon we have the need to. What you did here I can't say it's non acceptable or there are reasons to reject the approach right away, but I think we can use current issue to help you write self written code and understand to when and how to use third party code. This can be great reference when you will create this file on your own.

Here are couple of summary notes for current file:
- Most of the comments in this file are optional and rather distracting, I assume they used in the tutorial you have mentioned, to discribe each step they take, rather than how CSS work.
- There is lot to simplify in this code, but I feel like we will end up simplifying code of a tutor and miss nuances that are good for you to understand CSS when you face that problems and solve on your own.
- This more looks like a reusable menu component, but we don't plan to use it anywhere else yet, just in header, so I suggest to make it generic only when we understand how generic we suppose to make it to simplify the code and make it more semantic.
- It's not obvious what are you trying to do - adapt current menu into the existing design of CBA website, or adapt eventually website to match the style of the menu.

I'll suggest to have a screen-sharing or a pair programming session where we can first discuss what was your approach and what have you planned to do with current code and implementation, as I know it's just WIP and we can either discuss how we can adapt current code, or start from the scratch on an example of how to start developing the menu.

config.js
```diff
@@ -1,7 +1,8 @@
 const templateData =
 {
+
   site: {
-    title: "CMintS Starter", // See "layouts/_head.ejs"
+    title: "", // See "layouts/_head.ejs"
```
Manvel: Let's either delete current line, or use "CBA" instead, as it's used as a prefix on all page in chrome-automation.com.

config.js
```diff
@@ -1,22 +1,229 @@
...
   site: {
-    title: "CMintS Starter", // See "layouts/_head.ejs"
+    title: "CBA", // See "layouts/_head.ejs"
     navigation: [
       {
-        title: "CMintS",
-        link: "https://cmints.io/"
+        title: "Home",
+        link: "#",
+        sub_items:[]
```
Manvel: Suggestion/Detail: Let's remove empty subItems, as we don't seem to be needing those.

theme/layouts/_head.ejs
```diff
@@ -24,3 +24,7 @@
 
 <!-- Favicon -->
 <link rel="shortcut icon" href="<%= site.root %>/images/favicon.png">
+
```
Manvel: Detail: I did notice that most of the EJS files have 4 empty lines in the end. Let's keep it either to 1 empty line in the end of each file, or remove them completely so we can keep them consistent in future.

theme/layouts/_navigation.ejs
```diff
@@ -1,9 +1,21 @@
...
+<nav id="main-navigation">
+    
+  <% function recNav(navig) { %>
+    <ul> 
+      <% for ({title, link, sub_items} of navig) { %>
+        
+        <li>
+            <a href="<%- link %>" target="_blank"><%- title %> </a>
+         
+            <% recNav(sub_items)   %>
```
Manvel: Great job with the recursion, it mostly looks good, besides that we keep generating empty `ul` elements for the menu items that are missing `sub_items`. in order to stop generating redundant HTML content, let's have a condition here.

```
if (sub_items)
  recNav(sub_items)
```

config.js
```diff
@@ -1,22 +1,229 @@
...
+                  },
+                  {
+                    title: "Version 5.0", 
+                    link:"",
+                    sub_items:[]
+                  }
+        ]
+      }
+        ]
       }
```
Manvel: A general consistency suggestion: It's a good practice to keep the code consistent, though I know how hard it can sometimes be and that might become that frustrating little detail which might annoy a lot.

I did notice several inconsistencies in the object:
- inconsistent empty lines after each item.
- inconsistent space between opening and closing brackets.

There are programs that allow keeping those consistencies, for the JavaScript files one is called "ESlint", if we agree on styles we can use those to help us noticing such inconsistencies. We can talk about them in our meeting, so I can introduce you to those tools and as soon we agree on specific formatting we can setup some tools that will simplify future development. I'll leave some links below you might want to have a look.

References:
- Google style guide -> https://google.github.io/styleguide/
- Firefox style guide -> https://firefox-source-docs.mozilla.org/code-quality/coding-style/index.html
- Prettier -> https://prettier.io/
- Code linter -> https://eslint.org/

theme/less/_navigation.less
```diff
@@ -1,26 +1,109 @@
...
+#main-navigation {
+  background: linear-gradient(#fcfcfc, #ececec);
 }
+
+#main-navigation > ul
+{
+  display: flex;
+  padding: @listpadd;
+  height: 45px;
+  border-color: #EAEAEA #F5F5F5 #F5F5F5 #F5F5F5;
```
Manvel: Suggestion: I can't see that `EAEAEA ` is much different from `F5F5F5 `, so I wouldn't mind to use only `F5F5F5 ` to simplify this style, same applies to the other border styles below.

```
border: #F5F5F5 solid  1px;
border-radius: 5px;
```

See -> https://developer.mozilla.org/en-US/docs/Web/CSS/Shorthand_properties#Border_Properties

theme/less/_navigation.less
```diff
@@ -1,26 +1,109 @@
...
-  background-color: @secondary;
+
+#main-navigation {
+  background: linear-gradient(#fcfcfc, #ececec);
 }
+
+#main-navigation > ul
+{
+  display: flex;
+  padding: @listpadd;
```
Manvel: I'm not currently sure why are you using `@listpadd` here ? Why you think we will need to add a padding to top most `ul` item?

theme/less/_navigation.less
```diff
@@ -1,26 +1,109 @@
...
+  height: 45px;
```
Manvel: Why do we need to specify fixed height here?

theme/less/_navigation.less
```diff
@@ -1,26 +1,109 @@
...
+}
+
+#main-navigation ul
+{
+  list-style: none;
+}
+
+#main-navigation > ul ul
+{
+  top: @listpadd + 45;
```
Manvel: Seems like `45` is the height of the main navigation item, let's then reuse that value.

theme/less/_navigation.less
```diff
@@ -1,26 +1,109 @@
...
+  color: #595959;
+  
+}
+
+#main-navigation > ul > li > a
+{
+  display:block;
+  text-decoration: none;
+  font-weight: bold;
+  font-family: "PT Sans", "HelveticaNeue", "Helvetica Neue", Helvetica, Arial, sans-serif;
```
Manvel: Enhancement: Let's use website font's instead and remove this line for simplification.

theme/less/_variables.less
```diff
@@ -1,4 +1,5 @@
 // Colors
-@primary: #728448;
+@primary: white;
```
Manvel: I think instead of switching website's primary color to `white`, you wanted to remove background color from the <header> element.

theme/less/_variables.less
```diff
@@ -1,4 +1,5 @@
...
 @secondary: #52544e;
-@reverse: #fff;
+@reverse: black;
```
Manvel: Black is the default color, I don't think that we yet need this variable, we might add some meaningful ones as soon we need to. So I think instead you wanted to remove color attribute from `header > div ` element.

theme/less/_navigation.less
```diff
@@ -1,26 +1,109 @@
...
+#main-navigation ul ul > li > a
+{
+  text-decoration: none;
+  color: #595959;
+  
+}
+
+#main-navigation > ul > li > a
+{
+  display:block;
```
Manvel: Suggestion/personal-preference: I think they have used this to have effect of `vertical-align: top;`, let's use `vertical-align: top;` instead of switching to the display, block, I'm not sure if this behavior will be same on all browsers.

theme/less/_navigation.less
```diff
@@ -1,26 +1,109 @@
...
+}
+
+#main-navigation ul ul li:hover ul
+{
+  display: block;
+  position: absolute;
+  left: 99%;
+  top: 0;
+  margin-left: 1px;
+}
```
Manvel: Suggestion: Let's simplify this, for example we can also write:

```
#main-navigation > ul ul
{
  display: block;
  position: absolute;
  left: 0;
  padding: 0px;
}

#main-navigation > ul ul li ul
{
  left: 99%;
  top: 0;
  margin-left: 1px;
}

#main-navigation > ul li:not(:hover) > ul
{
  display:none;
}
```

This way you can also merge another style above `#main-navigation > ul ul` making the CSS smaller and simpler to maintain in future.

theme/less/_navigation.less
```diff
@@ -1,26 +1,109 @@
...
+  color:#545454;
+}
+
+#main-navigation ul ul li:hover
+{
+  background-color: #ffffff;
+  
+}
+
+//not taking effect when merging above and below selectors
```
Manvel: Yes, exactly, I think one below is redundant, because it adds background to the anchor element, which is expected to have no effect because it has transparent background by default, setting background color for parent item is completely fine.

theme/less/_navigation.less
```diff
@@ -1,26 +1,109 @@
...
+  top: 0;
+  margin-left: 1px;
+}
+
+#main-navigation > ul > li li, ul :hover
+{
+  background-color: #f0f0f0;
+}
+
+#main-navigation > ul > li > a:hover
```
Manvel: Suggestion: Let's move `#main-navigation > ul > li > a` style below above this style, or vice-versa, as those apply to the same element, and the difference is just in the "hover" state.

theme/less/_navigation.less
```diff
@@ -1,26 +1,109 @@
...
+
+}
+#main-navigation ul ul li
+{
+  padding:10px;
+  min-width: 150px;
+}
+
+
+#main-navigation ul ul > li > a
```
Manvel: Suggestion: we can simplify this selector and use instead `#main-navigation ul ul a` considering the fact that all our "a" tags are inside of the `<li>`s from that point. This way we have 1 level less nesting in the selector.

theme/less/main.less
```diff
@@ -4,20 +4,46 @@
...
+  text-decoration: none;
+  font-weight: bold;
+  line-height: 1px;
+  letter-spacing: -2px;
+  text-shadow: 0px 1px 0px #FFF;
+}
+
+header > div > h1 a:hover
+{
+  color:#3253ad;
```
Manvel: Suggestion: Let's simplify selectors here, you could use `header h3`, `header  h1` and `header  h1 a` instead.

theme/layouts/default.ejs
```diff
@@ -6,10 +6,11 @@
   <body>
     <header>
       <div>
```
Manvel: Suggestion:  outer `<div>` doesn't seem to be necessary here, let's remove it, to simplify our code unless it's here for a reason.

theme/layouts/default.ejs
```diff
@@ -6,10 +6,11 @@
...
-        <h1><%= page.header %></h1>
-      </div>
-    </header>
-    <% include _navigation %>
+        <h1><a href="#"><%= page.header %></a></h1>
```
Manvel: This's an interesting example of making a decision about semantic elements.

- `CBA Chromium Browser Automation (extension for chrome browser automation)` is the main title of the page
```
<h1><a><%= page.header %></a> <%= page.slogan %></h1>
```

- `CBA` is the main title of the page and `Chromium Browser Automation (extension for chrome browser automation)` is a paragraph text there.
```
<h1><a><%= page.header %></a></h1>
<p><%= page.slogan %></p>
```

- - `CBA` is the main title of the page and `Chromium Browser Automation (extension for chrome browser automation)` is the subtitle of the page.
```
<h1><a><%= page.header %></a></h1>
<h2><%= page.slogan %></h2>
```

It's tricky to make a choice here, but let's go with second suggestion here, unless you have a preference to other options and we can revisit that later if we decide it can have a SEO implications:
```
<h1><a><%= page.header %></a></h1>
<p><%= page.slogan %></p>
```

Anyway having `h3` after `h1` here is not what we need I guess.

theme/layouts/default.ejs
```diff
@@ -6,10 +6,11 @@
...
+        <h3><%= page.slogan %></h3>
```
Manvel: Suggestion: Let's define those variables in `config.js` and use `<%= site.title %>` and `<%= site.subtitle %>` instead.

Manvel (commented):
@naarrek Jan this looks much more better and we are on the right track, I have mostly added suggestions which should be easy to fix. You did a great Job here!

theme/less/_navigation.less
naarrek: I think we don't need here `display:block;` at all.

theme/less/_navigation.less
naarrek: Th website font is PT Sans. I'll leave it there.

theme/less/_variables.less
naarrek: ok

config.js
naarrek: OK

theme/layouts/_head.ejs
naarrek: OK

theme/layouts/_navigation.ejs
naarrek: OK

theme/less/_navigation.less
naarrek: Let's talk about this comment on our meeting.

theme/less/_navigation.less
naarrek: Removed.

theme/less/_navigation.less
naarrek: OK

theme/less/_variables.less
naarrek: The alternative option is to have no background-color in header section. OK

theme/less/_navigation.less
naarrek: OK

theme/less/_navigation.less
naarrek: The problem is resolved :)

theme/less/_navigation.less
naarrek: Don't really found the second style.

theme/less/_navigation.less
naarrek: OK

theme/less/main.less
naarrek: OK

theme/layouts/default.ejs
naarrek: OK

theme/layouts/default.ejs
naarrek: OK. Done

theme/layouts/default.ejs
naarrek: OK

config.js
naarrek: Agree, let's talk about ahead on meeting.

theme/less/_navigation.less
Manvel: As was discussed in the meeting, we don't need `@listpadd;` variable, as well as the `height` property below.

config.js
```diff
@@ -133,132 +133,6 @@ const templateData =
...
-              {
-                title: "bg-inject",
-                link: ""
-              },
-              {
-                title: "bg-function (deprecated)",
-                link: ""
-              },
-              {
-                title: "change",
```
naarrek: Eslint show here indentation problem, so it will be fixed.

config.js
Manvel: As the linters are introduced for most of the part of the code, resolving this comment.

pages/index.ejs
```diff
@@ -1,7 +1,6 @@
 ---
 title: Homepage
-description: Get your webiste up and running online in no time using current starter
-header: CMintS starter
+description: CBA Chromium Browser Automation (extension for chrome browser automation)
```
Manvel: Detail: Seems like description is not being used per page basis, we might eventually introduce description to improve SEO, but as it's not a requirement and as we are using `subtitle` variable specified in the `config.js` to show the sub header we can either remove this or return it to it's initial state.

theme/layouts/_navigation.ejs
```diff
@@ -1,9 +1,25 @@
-<nav>
-  <ul>
-    <% for (const {title, link} of site.navigation) {%>
-      <li>
-        <a href="<%- link %>" target="_blank"><%- title %></a>
-      </li>
-    <% } %>
-  </ul>
+<select id="mobile-nav">
+      <% for ({title, link, sub_items} of site.navigation) { %>
```
Manvel, line 2: Detail/suggestion: Unfortunately there wasn't a linter and code formatter that would support EJS files, so for now let's try keeping the file consistent as much as we can, for example here we can introduce 2 level indentation instead of 6, to make the code more compact and consistent, same applies to other inconsistent identations in the file . It should be easy to setup identation in the VSCode -> https://stackoverflow.com/a/38556923/1439474, so when you hit TAB it will use "2 spaces".

Manvel: Detail/suggestion: Unfortunately there wasn't a linter and code formatter that would support EJS files, so for now let's try keeping the file consistent as much as we can, for example here we can introduce 2 level indentation instead of 6, to make the code more compact and consistent, same applies to other inconsistent identations in the file . It should be easy to setup identation in the VSCode -> https://stackoverflow.com/a/38556923/1439474, so when you hit TAB it will use "2 spaces".

theme/layouts/_navigation.ejs
```diff
@@ -1,9 +1,25 @@
...
+        <option value="<%- title %>"><%- title %></option>
+      <% } %>
+</select>
+
+<nav id="main-navigation">
+  <% function recNav(navig) { %>
+    <ul> 
+      <% for ({title, link, sub_items} of navig) { %>
+
```
Manvel: Detail: let's remove this trailing new line, in order to make this part compatible with other similar pieces in current file.

theme/layouts/_navigation.ejs
```diff
@@ -1,9 +1,25 @@
...
+        <li>
+            <a href="<%- link %>" target="_blank"><%- title %> </a>
+              <% if (sub_items == null) {
```
Manvel: This probably can be simplified, as you don't have to check here whether the item doesn't exist, also note that you need to close the tag before continuing to the next action, ex.:
```js
<li>
  <a href="<%= link %>" target="_blank"><%= title %></a>
    <% if (sub_items) { %>
      <img src = '../images/arrow_bullet.png'/>
      <% recNav(sub_items);
    } %>
</li>
```

Manvel: Also here is an interesting thing about `<%=` and `<%-` and security in general.
- `<%=` - is escaping HTML tags.
- `<%-` - doesn't not escape HTML tags.

It's always better to use `<%=` where possible, because even though we currently do control our inputs this eventually might become not true and for example not escaped HTML tags can lead to the code injections and [XSS attacks](https://en.wikipedia.org/wiki/Cross-site_scripting).

As a simple example, imagine when using a commenting service and loading user comments here, in which case we wouldn't control inputs, in that case the comment for example could contain `<script></script>` tag with a malicious code. So even though in our case it might be safe to use, it's always a good practice to try keep the code secure from the beginning where possible. Or for a better example consider a case when we start allowing users to edit our pages similar to wiki service and detail like this can eventually become harmful.

This scenarios much likely wouldn't happen, as we currently plan to control our input flow, but still a lot of famous vulnerabilities has evolved from similar assumptions.

theme/layouts/_navigation.ejs
```diff
@@ -1,9 +1,25 @@
...
+                continue;
+                } else { %>
+                  <img src = '../images/arrow_bullet.png'/>
+                  <%recNav(sub_items)  ; 
+                }%>
```
Manvel: Detail: let's remove trailing space here and on other lines where those exist.

theme/less/_navigation.less
Manvel: This comment is still valid.

theme/less/_variables.less
Manvel: This comment is still actual.

theme/less/_navigation.less
```diff
@@ -1,25 +1,103 @@
...
-  ul
+  background: linear-gradient(#fcfcfc, #ececec);
+}
+
+#main-navigation > ul
+{
+  height:50px;
+  display: flex;
+  padding: 0;
+  border-color: #EAEAEA;
```
Manvel: As mentioned here https://github.com/browser-automation/cba-website/pull/7#discussion_r404002510 

We can also simplify the border rules here:
from:
```
border-color: #EAEAEA;
border-radius: 5px;
border-style: solid;
border-width: 1px;
```
to:

```
border: #F5F5F5 solid  1px;
border-radius: 5px;
```

theme/layouts/default.ejs
```diff
@@ -5,11 +5,10 @@
   </head>
   <body>
     <header>
-      <div>
-        <h1><%= page.header %></h1>
-      </div>
-    </header>
-    <% include _navigation %>
+        <h1><a href="#"><%= site.title %></a></h1>
```
Manvel: Detail: let's use consistent indentation here with parent tags, 2 spaces, instead of 4.

theme/layouts/default.ejs
```diff
@@ -5,11 +5,10 @@
...
+        <p><%= site.subtitle %></p>
+        <% include _navigation %>
```
Manvel: Detail: Let's remove trailing spaces where possible on EJS files.

theme/layouts/_navigation.ejs
```diff
@@ -1,9 +1,25 @@
...
-  <ul>
-    <% for (const {title, link} of site.navigation) {%>
-      <li>
-        <a href="<%- link %>" target="_blank"><%- title %></a>
-      </li>
-    <% } %>
-  </ul>
+<select id="mobile-nav">
+      <% for ({title, link, sub_items} of site.navigation) { %>
+        <option value="<%- title %>"><%- title %></option>
```
Manvel: Here we only list top level items, while in the Drupal website subitems are also listed here.

theme/layouts/_navigation.ejs
```diff
@@ -1,9 +1,25 @@
-<nav>
-  <ul>
-    <% for (const {title, link} of site.navigation) {%>
-      <li>
-        <a href="<%- link %>" target="_blank"><%- title %></a>
-      </li>
-    <% } %>
-  </ul>
+<select id="mobile-nav">
```
Manvel: That's another tricky and interesting part for designing for a semantic web which we will improve later on.

Mobile navigation is still a navigation item and we either need to move it into our existing `<nav>` item below, or nest it into a new `<nav>` item. Both options has it's pros and cons from semantic and designing point of view. And I'm having hard time deciding for one or another right now. So whichever option you think is best please go for it, we might revisit this when we start to enhance our menu items.

theme/less/_navigation.less
```diff
@@ -1,25 +1,103 @@
...
-        color: @reverse;
-      }
-      display: inline-block;
-      padding: 10px 10px 10px 0;
-    }
-    padding: 0 30px;
-    list-style: none;
-    max-width: 1000px;
-    margin: 0px auto;
+    display: inline-block;
```
Manvel: Suggestion/detail: We can also use `display: block;` here because we are not going to use features of `inline-block` here other, than it acting as a block level element.

theme/less/_navigation.less
```diff
@@ -1,25 +1,103 @@
...
+
+#main-navigation 
 {
-  ul
+  background: linear-gradient(#fcfcfc, #ececec);
+}
+
+#main-navigation > ul
+{
+  height:50px;
```
Manvel: Detail: Let's either reuse this height for `top ` in the `#main-navigation > ul ul` selector or remove this here, I would rather reuse it, because it will prevent errors of hardcoding hack we have discussed.

Consdering that we are not planing to apply this property to whole website, like general colors which we keep in `variables.less`. You can try making use of [CSS variables](https://developer.mozilla.org/en-US/docs/Web/CSS/var).

In this case we might have something like:
```
#main-navigation
{
  --menu-height: 50px;
}
```
And use it by `height: var(--menu-height);`

Note: When/if you move the mobile code into current `<nav>` the example might need to be updated.

theme/less/main.less
```diff
@@ -4,20 +4,37 @@
...
 {
-  margin: 0px auto;
-  max-width: 1000px;
-  color: @reverse;
+  color: #7d7d7d;
+  font-size: 24px;
+  letter-spacing: -1px;
+}
+
+header h1, header h1 a
```
Manvel: Suggestion: We do apply duplicate styles here, while in some cases it would be good idea, here we are missing some optimization opportunities. we can either use LESS here, or move `header h1 a ` selector from here:
```
header h1 a
{
  color: #262626;
  text-decoration: none;
}
```

After we can also notice that there are some common styles in whole header anchor elements, if needed and you want we can also improve further and have:
```
header a
{
  text-decoration: none;
}
```

That way we don't need to specify that rule anywhere else in the header and simplify our selectors, just a small suggestion.

Manvel (commented):
@naarrek Now the code looks much more better, I have made further suggestions, most of them because we don't have a linter solution for `EJS` files :( If you find those too frustraighting, let's discuss further how we can simplify the development process, meanwhile I'll think about solution that will automate some common patterns as soon we find those.

Other than that there are some suggestions and reminders which probably have been lost in the noise.

theme/less/_navigation.less
```diff
@@ -1,25 +1,103 @@
...
+{
+  text-decoration: none;
+  font-weight: bold;
+  font-family: "PT Sans", "HelveticaNeue", "Helvetica Neue", Helvetica, Arial, sans-serif;
+  font-size: 14px;
+  color:#2883DE;
+  text-shadow: 1px 1px 0px #fff;
+}
+
+#main-navigation ul li img
```
Manvel: Suggestion: we can also use `::after` to simplify the markup and supposedly we shouldn't need to use `` tag here as it doesn't provide any accessibility feature here.

Manvel: Also not sure if `float: right` is required here.

Manvel: Also, there are 2 states in the https://chrome-automation.com/ there is a hover as well. I think most simple way would be to use `::after` here, note that [:has](https://developer.mozilla.org/en-US/docs/Web/CSS/:has) is not yet supported by major browsers, so we might need to assign a classname to the `li` element.

naarrek: We can not select which menu entry has submenu. Thus we need at least a non-empty `span` to add background image and select the `span` for other properties.

theme/layouts/_navigation.ejs
naarrek: Done.

theme/less/_variables.less
naarrek: Somehow I did not managed to imply it in this commit but it was fixed by removing the `background-color` property.

theme/less/_navigation.less
naarrek: Which are website fonts?

theme/less/_variables.less
Manvel: Fare enough, but we now have variable that is not used anywhere in the code. I suggest one of those:
- Remove variable completely.
- Use more meaning-full color value and reuse where applicable.
- Revert the color to it's initial state `#728448` to provide less noise in the review and take care of it in the future.

I lean towards options 1 and 2, because we will need to create a followup issue for 3 to keep track of what needs to be done.

theme/less/main.less
naarrek: Good option! Done.

theme/less/_navigation.less
Manvel: Currently the [way they are specified here](https://github.com/browser-automation/cba-website/pull/7/files#diff-ae30bf8ab3049325091a23f3bf9aaa4aR7) is same fonts as here, so you can remove current property, as the `font-family` styles are [inherited](https://developer.mozilla.org/en-US/docs/Web/CSS/inheritance), the same font will be applied here. So basically what I mean is that you just need to remove current line, that way when we change font specified on `<body>` that will affect whole website including menu items.

theme/less/_navigation.less
Manvel: As I have mentioned in the comment, you can use a classname instead.
Maybe something like [in layout code](https://github.com/browser-automation/cba-website/pull/7/files#diff-18f055c29e5211ab0d148460aa1716e9R13):
```
<li <%= sub_item ? "class="has-children" : "" %>
```

And use 
`#main-navigation ul li:sub_item::after` or `#main-navigation ul li:sub_item a::after`

As an example.

theme/layouts/_navigation.ejs
```diff
@@ -1,9 +1,25 @@
...
+      <% } %>
+</select>
+
+<nav id="main-navigation">
+  <% function recNav(navig) { %>
+    <ul> 
+      <% for ({title, link, sub_items} of navig) { %>
+
+        <li>
+            <a href="<%- link %>" target="_blank"><%- title %> </a>
```
Manvel: Seems like in the Drupal website links were not using `target="_blank"`, this will open links in the new tab.

theme/less/_navigation.less
```diff
@@ -1,25 +1,103 @@
...
+  border-color: #EAEAEA;
+  border-radius: 5px;
+  border-style: solid;
+  border-width: 1px;
+  box-shadow: 0 1px 1px rgba(0, 0, 0, 0.2);
+}
+
+#main-navigation > ul li
+{
+  padding: 15px;
```
Manvel: We might need to use this padding on the anchor tag, because currently the link is only working when you hover on the text, but not on the menu item, same applies to the padding of subitem li element, which might also need to be moved into their anchor tag accordingly.

theme/layouts/_navigation.ejs
naarrek: > 
> 
> This probably can be simplified, as you don't have to check here whether the item doesn't exist, also note that you need to close the tag before continuing to the next action, ex.:
> 
> ```js
>     <% if (sub_items) { %>
> ```

This will throw error `navig is not iterable` dont really know why.
Besides that we should use here `continue` to make loop jump to the next iteration if there is no sub_items.

Manvel: > we should use here continue to make loop jump to the next iteration

Having loop here, will prevent invocation of rest of the code and result in unclosed `<li>` tag, the error went unnoticed, because modern browsers are smart enough to prevent user from seeing markup error. You can observe that by clicking `view page source` in Chrome. You might see :
```
--
  <nav id="main-navigation">
  
    <ul> 
      

        <li>
            <a href="#" target="_blank">Home </a>
              

        <li>
            <a href="#" target="_blank">Documentation </a>
              
                  <img src = '../images/arrow_bullet.png'/>
                  
    <ul> 
      

        <li>
            <a href="" target="_blank">Interface </a>
              
                  <img src = '../images/arrow_bullet.png'/>
                  
    <ul> 
```
That's because of `continue` here, as it prevent the invocation of the rest of the code inside for loop. for loop will continue innovation for each item in the array irregardless of the `continue` unless there is a `return` or blocked code invocation.

Also another interesting note, you are checking `sub_items ` against, `null` and this check only passes because JavaScript is funny and `undefined == null` result in true, but `undefined === null`(check against the type) result in false. So in this case checking `if (sub_items)` may be  a better choice?

I can see some more errors in current implementation - `</ul>` being closed inside of the for loop, while being open outside.

`navig is not iterable ` more sounds like corrupted input, when `sub_items` passed to the recursion is not an array, that's an assumption, but seems like the code needs to be revisited in general considering the latest observation.

Suggestion: Ensure that the input in `config.js` is not corrupted, check small against the output and see if you get expected output or not, by observing generated HTML.

theme/less/_navigation.less
naarrek: Resolved

theme/layouts/_navigation.ejs
naarrek: Right, but I think subitems of top level items are only displaying when corresponding pages are open, so we need pages for these to show up.

naarrek: Right, but I think subitems of top level items are only displaying when corresponding pages are open in drupal website, also the 3-rd level subitems don't show up at all. We must decide how to deal with that.

theme/less/_navigation.less
naarrek: Here I removed `height` option and left menu height to control padding of anchors. Is it Ok?

theme/less/_variables.less
naarrek: The issue of colors is not simple. we need to use less colors, in the other hand the colors of drupal site varies in close range and it is hard to choose which one is primary and secondary. I suggest to remove variables leaving the file empty and return to this issue when site and its colors will be more definite.

theme/less/_navigation.less
naarrek: I will use here anchor class. Because `li` is part of recurrent flow and moving it or adding additional `li` will cause recurrent function to work not correct I think.

theme/less/_variables.less
Manvel: Agree

theme/layouts/_navigation.ejs
Manvel: This comment is still relevant.

theme/layouts/default.ejs
Manvel: Still can see trailing spaces.

theme/layouts/_navigation.ejs
Manvel: As agreed, we will implement simple solution, either with 1 or 2 level.

theme/less/_navigation.less
Manvel: It's okey, we can revisit this issue, when we face a problem and/or during enhancements.

theme/less/_navigation.less
Manvel: Not sure if I understand your argument. Can you please elaborate on 
`Because li is part of recurrent flow and moving it or adding additional li will cause recurrent function to work not correct I think ` ?

What I can see you have just duplicated `<a>` tags in your markup:
1. This will eventually provide you much of headache when you decide to make your menu accessible, because both anchor elements are focusable, so you might need to make one not focusable which eventually will require you to create a hack to highlight whole menu item.
2. It's not necessary to duplicate those anchors, not sure how is this different/better from using `<span>` as you have done previously.

Let's talk about this issue, to understand what bother you about using solution suggested above and try to understand all pros and cons.

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
Manvel: This is VSCode specific file, you can add `.vscode` directory to the `.gitignore`, so we won't track it anymore.

It's quite handy as you won't need to untrack those when using `git add .`.

how REVISION
```diff
@@ -0,0 +1,76 @@
+[33mcommit ce194e677659a9f500a7f0affc8957d236918f35[m[33m ([m[1;36mHEAD -> [m[1;32missue_5[m[33m)[m
```
Manvel: I guess this file ended up here by mistake.

theme/layouts/_navigation.ejs
```diff
@@ -1,9 +1,24 @@
...
+
 <nav>
-  <ul>
-    <% for (const {title, link} of site.navigation) {%>
-      <li>
-        <a href="<%- link %>" target="_blank"><%- title %></a>
-      </li>
-    <% } %>
-  </ul>
+<select id="mobile-nav">
```
Manvel: Seems like the JavaScript part is yet missing.

In order to make those items navigatable, you need to listen to onchage event:
Here is [an example used in cmints.io](https://github.com/cmints/website/blob/master/theme/js/main.js)

theme/layouts/_navigation.ejs
```diff
@@ -1,9 +1,24 @@
...
+  <% for ({title, link, sub_items} of site.navigation) { %>
+    <option value="<%= title %>"><%= title %></option>
```
Manvel: Guess you meant to use link of the item as the value

theme/layouts/_navigation.ejs
```diff
@@ -1,9 +1,24 @@
+<nav id="main-navigation">
+  <% function recNav(navig) { %>
+    <ul> 
+      <% for ({title, link, sub_items} of navig) { %>
+        <li>
+            <a href="<%= link %>"><%= title %></a>
+              <% if (sub_items) { %>
+                <a class="has-children" href="<%= link %>"></a>
```
Manvel: Detail/suggestion: Reading the classname, it sounds like `<a>` tag has children, while initially this classname meant to be used on the `<li>` element which technically can have children.

Manvel: As mentioned in https://github.com/browser-automation/cba-website/pull/7#discussion_r409176587 this still looks like a markup hack to visually show arrow element, though previously(with span) it has fewer impact on accessibility.

theme/less/_navigation.less
```diff
@@ -1,25 +1,114 @@
...
+  font-weight: bold;
+  font-size: 14px;
+  color:#2883DE;
+  text-shadow: 1px 1px 0px #fff;
+}
+
+#main-navigation .has-children
+{
+  background: url('../images/arrow_bullet.png') no-repeat 90% 50%;
+  float: right;
```
Manvel: I can see that, you have used flexbox above for this element, that supposedly should make the "previous way" of using floats obsolete?

theme/less/_navigation.less
```diff
@@ -1,25 +1,114 @@
...
+  margin-right: 10px;
+}
+
+#main-navigation > ul > li:hover > a.has-children
```
Manvel: Detail: Seems like `#main-navigation > ul > li:hover` also uses difference background gradient color in Drupal website.

theme/less/_navigation.less
```diff
@@ -1,25 +1,114 @@
...
+{
+  background: url('../images/arrow_bullet_dark.png') no-repeat 90% 50%;
```
Manvel: "../" means upper directory, while this works on homepage, it might not work on other pages. It's better to use absolute path starting with `/` instead of relative for the resources. Same applies to the image below.

Manvel (commented):
Well done Narek jan with fixing the recursion issues, another round of the review is done.

Small suggestions besides of arrow issue suggestion of using `::after` selector instead of HTML markup hack and missing JavaScript code for the pages navigation.

how REVISION
naarrek: Not sure by which means it has been created.

.vscode/settings.json
naarrek: Done. But don't know how it appeared here, i didn't add it to commit.

theme/less/_navigation.less
naarrek: Right, this shouldn't be there.

naarrek: Right, this shouldn't be there

theme/less/_navigation.less
Manvel: `#F0EEF0`

.vscode/settings.json
Manvel: Maybe when adding all files in the directory, example by `git add .` which adds all file including untracked files into commit staging. It's easy to miss that, in my case a lot of files related to my system used to be added and some of them already supposedly should be listed in `.gitignore`.

how REVISION
Manvel: Thinking, maybe by piping output of  git log into a file, ex.: `git log -3 > REVISION`

.gitignore
```diff
@@ -1,4 +1,6 @@
 node_modules
 public/css/main.css
 public/js/main.js
-package-lock.json
\ No newline at end of file
+package-lock.json
+settings.json
+how REVISION
```
Manvel: Why have you added those files to here?

Usually you only add those files and documents to `.gitignore` that are created during the build time, automatically by the software you are using, but you don't have to add every single file that you have created for a reason.

.vscode/settings.json
Manvel: This comment is still valid.

how REVISION
Manvel: This comment is still valid.

theme/layouts/_navigation.ejs
```diff
@@ -1,9 +1,23 @@
...
+
 <nav>
-  <ul>
-    <% for (const {title, link} of site.navigation) {%>
-      <li>
-        <a href="<%- link %>" target="_blank"><%- title %></a>
-      </li>
-    <% } %>
-  </ul>
+<select id="mobile-nav">
```
Manvel: Detail: this is bit inconsistent with the above declaration - `<nav id="main-navigation">`, above we used "mobile-nav" on the navigation item, while here it's the select item, in order to make markup more readable and consistent, I'll suggest to use "mobile-nav" on the parent `<nav>` element instead or change the ID name here. Though I think having ID specified on the nav item, might be useful in future as well and provide more flexibility, as you will use the issue of differentiating between 2 different navigation items.

theme/js/main.js
```diff
@@ -1,3 +1,11 @@
 const {message} = require("./_module");
 
 console.log(message);
```
Manvel: Let's remove irrelevant code here. It was for demonstration purposes only.

theme/js/main.js
```diff
@@ -1,3 +1,11 @@
...
+
+// Mobile Menu
+const mobMenu = document.querySelector("#mobile-nav");
+mobMenu.addEventListener("change", e =>
+{
+  const data = document.querySelector(".data");
+  data.textContent = `Data to be added here.`;
```
Manvel: I think when we start learning JavaScript, things like this will become more intuitive for you. Currently this code doesn't make a redirection. This code says "Select element with class `.data` and add `Data to be added here.` string as it's text content". I think here you want to find out which item from the `<select>` element was selected by user and navigate to that specific page. As [was mentioned here](https://github.com/browser-automation/cba-website/pull/7/files#r409189013) you could check a similar example.

You might want here to do the following after the [change event](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/change_event) has been fired:
- Get the selectbox element
- Find out the changed value of the selectbox.
- Navigate to the selected page according to the changed value.

First you need to get the selectbox element, in fact you have it defined above in the line: `const mobMenu = document.querySelector("#mobile-nav"); `, where you have used [querySelector](https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelector) method which can help you find an element using CSS selector.

You can then use [select element's value attribute](https://html.spec.whatwg.org/multipage/form-elements.html#dom-select-value) which returns value of the first selected options element, as we don't have multiselect selectbox here, we can rely on that result. After you have the value, the only thing is to navigate to the page specified in the value.

In order to navigate the page according to the user selection in the selectbox we can use [window](https://developer.mozilla.org/en-US/docs/Web/API/Window) object's [location](https://developer.mozilla.org/en-US/docs/Web/API/Window/location) property. In the example I have posted above, we have used that property to create an absolute URL, from protocol, domain and pathname, but the thing is that we can do it even more simpler by modifying [window.location.pahtname](https://developer.mozilla.org/en-US/docs/Web/API/Location/pathname), which actually contain the pathname of the page you are located in, by changing the pathname you will also make a redirection.

theme/layouts/_navigation.ejs
```diff
@@ -1,9 +1,23 @@
+<nav id="main-navigation">
+  <% function recNav(navig) { %>
+    <ul>
+      <% for ({title, link, sub_items} of navig) { %>
+        <li <%- sub_items ? 'class="has-children"' : '' %>>
```
Manvel: Detail/consistency: In the current code we are using double quotation marks for the string literals, let's keep it consistent, one way of the other, so you can write `<%- sub_items ? "class='has-children'" : "" %>` instead.

.gitignore
Manvel: I think you meant to add here `.vscode` folder, which keeps generating your local settings in `.vscode/settings.json` and probably also some other local files.

Manvel (commented):
Had a look into the latest code you have added, but haven't performed full code review yet.

theme/layouts/_navigation.ejs
Manvel: This comment is still relevant.

theme/layouts/default.ejs
Manvel: This comment is still relevant.

theme/layouts/_navigation.ejs
Manvel: This comment is still relevant.

Update: My bad, forgot to pull the latest changes. this is done, thanks.

theme/less/_navigation.less
```diff
@@ -1,25 +1,132 @@
...
+
+#main-navigation > ul > li.has-children > a:hover:after
+{
+  content: url('/images/arrow_bullet_dark.png');
+  padding-left: 10px;
+}
+
+#main-navigation ul ul li.has-children > a:hover:after
+{
+  content: url('/images/arrow_bullet_dark.png');
```
Manvel: I'm impressed by this! Good job. Didn't know that it's possible to use `url()` in content.

theme/js/main.js
Manvel: Also the `./_module` file can be removed I guess.

theme/layouts/_navigation.ejs
Manvel: BTW: If you feel like using `'` instead of `"` I'm happy to switch to.

theme/layouts/_navigation.ejs
naarrek: You mean in general? I removed this part of code.

Manvel: Yes in general, but I can see that on this line(line below `recNav()` call), there are trailing spaces also now, by trailing spaces I mean spaces in the end of the line:

Screen Shot 2020-04-19 at 11 58 04 PM

how REVISION
naarrek: I

> 
> 
> Thinking, maybe by piping output of git log into a file, ex.: `git log -3 > REVISION`

Then if we don't need this file I'll just remove it.

.vscode/settings.json
naarrek: > 
> 
> This comment is still valid.

When I was committing git status didn't show anything this time. Anyway, I corrected my mistake and added .vscode folder to .gitignore instead of settings file. Hope it will resolve this.

theme/layouts/default.ejs
naarrek: Done

.vscode/settings.json
Manvel: Yes, it didn't because you already had the file in previous commits, when adding file and/or directory to the `.gitignore` it doesn't automatically hint you that you might want to remove the commited file. It basically prevent future files in the directory to be accidentally commited. So you basically want to manually delete the file, after updating `.gitignore` to ensure the old version and/or furhter modification, won't be pushed in future.

naarrek: > So you basically want to manually delete the file, after updating `.gitignore` to ensure the old version and/or furhter modification, won't be pushed in future.

In settings.json I thought there is useful property of `"editor.tabSize": 2`. Anyway I removed it and it has no effect on tabbing. It is still 2 spaces.

Manvel: I see your point, it is a useful property for your use-case, but there are multiple reasons why it's not a good idea to commit your local IDE settings:

1. As soon me/you/other contributor add any new setting in the local `.vscode` setting, they will still appear in the staging directory, so the person will end up adding new settings to the commit accidentally.
2. Those are VSCode related settings, but it doesn't necessarily mean that other contributors might use VSCode as well and that might eventually lead into maintaining multiple IDE specific settings.

I think what might have been not clear here is that after removing this file from the commit, you can still have your local settings in `.vscode` folder and/or have a global settings, unless you delete the folder yourself. So basically, after removing this file from the commit, please set that setting in your IDE again and they should stay there.

I would rather rely on the linters to pick up similar issue, rather than enforcing IDE specific rules. The problem is with [our linter](https://github.com/browser-automation/cba-website/issues/16), but not with the IDE.

It's a great idea of yours to have best practices for VSCode and/or other IDEs, as an idea we can have a [WIKI page for that with the recommended settings and/or general best practises](https://github.com/browser-automation/cba-website/wiki), or a new repository with the various VSCode settings, that are common for [browser-automation organization](https://github.com/browser-automation), but as we haven't identified a common repository best practices yet, wiki might be a good start.

theme/js/main.js
```diff
@@ -1,11 +1,8 @@
...
-
-console.log(message);
-
 // Mobile Menu
 const mobMenu = document.querySelector("#mobile-nav");
 mobMenu.addEventListener("change", e =>
 {
-  const data = document.querySelector(".data");
-  data.textContent = `Data to be added here.`;
+    const optionTitle = e.target.selectedOptions[0].label.split(' ').join('');
```
naarrek: I am aware of linting errors. I forgot to pass the tests. Let's leave it for the next commit.

theme/js/main.js
```diff
@@ -1,3 +1,6 @@
-const {message} = require("./_module");
-
-console.log(message);
+// Mobile Menu
+const mobMenu = document.querySelector("#mobile-nav");
```
Manvel: Suggestion: We don't need to listen to multiple elements inside  `#mobile-nav`, so we can instead use selectbox, i.e.: 
```
const mobMenu =  document.querySelector("#mobile-nav select");
...
window.location.pathname = mobMenu.selectedOptions[0].value;
```

theme/less/_navigation.less
```diff
@@ -1,25 +1,133 @@
...
+
+#main-navigation ul ul li:hover, ul ul a:hover
+{
+  background-color: #ffffff;
+}
+
+#main-navigation ul a
+{
+  padding: 15px;
+  flex-grow: 2;
```
Manvel: Detail: I assume this left since the times arrow element was sibling of anchor tag, so you have specified this to push the content all the way to the end.

Manvel (approved):
Hey Narek jan, good job finishing your first real world task.

I have left 2 comments, but as those are just suggestions and a detail that doesn't affect the view, I'll leave that up to you if you would like to address those before merging or comeback to them later on.

Feel free to merge the changes when you are ready.

Before merging please use a meaningful commit message ex.: "Issue #5 - Implemented header with drop down navigation menu", for that you can just update the PR Title and ensure that your commit message will be used when you use "Squash Merge", yes please use "Squash Merge" in order to merge 1 meaningful commit into the master branch, that way it will be simpler in future to navigate through history.

theme/less/_navigation.less
naarrek: Basically we can replace it, for example, by changing  padding (increasing padding-right component) of `#main-navigation ul a`, because there is a gap between 2 flex items (not covered by anchor selectors). But in that case we don't know exact size of the gap, flex-grow in this case is what we need I think, because it covers the unused space between flex items. 
By the way we can remove  `justify-content:space-between;` because this taking no effect on positioning flex items and is overwritten with properties of `:after` selector as it is positioned absolutely.

Manvel: I still don't understand, why you need `flex-grow` here?

All `#main-navigation ul a` item siblings as far as I can see are positioned absolutely, so there seem to be no item you need to push by `flex-grow` element. Am I missing something?

> By the way we can remove justify-content:space-between;

Sounds good, please remove all unused items, I may have missed that, but if you tested and see no effect please go ahead.

Issue #12, opened 2020-04-10
Import CBA website content


Manvel:
# Background

Page contents currently is available in the CSV format exported from the database of the Drupal website. The pages on the, pages consisted of Title and the actual content.

# What to change

Add pages to the website using the [exported content](https://docs.google.com/spreadsheets/d/1Bn6hMObKN6oRwiDUd5fKYx59pBnwyzo9Ex2bFc6yJMo/edit#gid=1368966783).

Manvel:
In order to support hierarchical structure of the navigations in the website, we can introduce the nested hierarchy in the pages folder to easier find the relevant content in the pages directory ex.:
```
|- documentation
|  |- interface
|     |- index.html
|     |- project.html
|     |- actions-grid.html
...  
```

Also we can provide hierarchy reference in the pages in order to construct navigation menus, ex.:
```
actions-grid.html
---
title: Actions datagrid
parentBranch: interface
---

...
```

```
documentation/interface/index.html
---
title: Interface
branch: interface
---
...
```

This way we will be able to use [queryPages](https://cmints.io/en/documentation/themes/ejs#site.querypages) to construct the navigations and breadcrumbs in theme folder, ex.:
```
<% 
  const docPages = site.queryPages((data) => data.branch || data.parentBranch); 
  for (const {branch, parentBranch} of docPages)
  ...
%>
```

PR #18, opened 2020-05-03
Issue 12

naarrek:
Setting conditions and tools for migration.


config.js
```diff
@@ -184,6 +184,7 @@ const templateData =
   }
 };
 
+dirTree = require("directory-tree");
```
Manvel: In general this shouldn't work, but the reason why such declaration made "dirTree" a global variable is that we are not using [strict mode](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode). In order to avoid that we can add a new [strict](https://eslint.org/docs/rules/strict) rule into our eslinter `strict: ["error", "global"]` seem to enforce that  and/or you can just add `"use strict";` on top of this file.

In general exposing variables into global scope is considered to be a bad practice unless that's really necessary, as there is a risk of miss-using global variables, or having unexpected behavior in local functions that are using same variable naming.

I think what you wanted here is to introduce something that is called EJS helpers, in order to do that you can use "templateData" object and pass the desired method as one of it key's values. Something like:
```js
const dirTree = require("directory-tree");
const templateData = {
  site: {
    dirTree,
    ...
  },
  ...
}
```

Then access it by `<% site.dirTree(..) %>;`, or what might be better, is to initially get the structure and pass it into the templateData, to abstract away details like what is ".page" outside of the EJS template and avoid rerunning that function each time you reuse your EJS template i.e.:
```js
const dirTree = require("directory-tree");

const templateData = {
  site: {
    tree: dirTree("./pages").children,
    ...
  },
  ...
}
```

theme/layouts/_navigation.ejs
```diff
@@ -1,17 +1,41 @@
...
-            <% if (sub_items) { %>
-              <% recNav(sub_items);
+      <% for (let {path, name, children, size, type} of ft) {
+            if (type == 'directory') {
+              excludeFile.push(name);
+              const ext = '.md';
+              path = path.concat(ext); %>
+              <li <%- children ? "class='has-children'" : "" %>>
+              <a href="<%= path %>"><%= name %></a>
+              <% if (children) {
```
Manvel: I'm not sure why are you checking for the directory and then for the children, the only case when directory will not have children is that when the directory is empty, but I believe that you only wanted to check if the item has children here? Though it's not really what you might want to do because there are some edge cases when index in web and in programming might have a special meaning as in this case when `mypage.md` and `mypage/index.md` are same page while in case of later one you will get type="directory" and children for "mypage" item, when in the other case no.

theme/layouts/_navigation.ejs
```diff
@@ -1,17 +1,41 @@
...
+              <% }
+      } %>
     </ul>
-  <% }
-  recNav(site.navigation) %>
+  <%
+  }
+  filteredTree = dirTree('./pages', '.md').children;
+  recNav(filteredTree);
+  %>
```
Manvel: Looking into this implementation, I feel like that I have failed to describe the task well enough. I know that you also noticed that something is off, in your implementation as CMintS might have already some solutions for you and you might not need a third-party tool. But first let's try to understand what is wrong with current implementation:
1. Clean code implemented in previous PR was replaced by more complex one for similar parts.
2. "directory-tree" creates different data-structure from what you expect in [initially created data-structure](https://github.com/browser-automation/cba-website/blob/master/config.js#L6).
3. We now construct item titles from filenames, instead of title of pages, so if I rename my file the menu item label will change.
4. You try to use folders structure to generate page path, but now we won't be able to use permalinks and there are some edge cases how "index" pages are treated.

Considering that this route causes more harm than help you solving the issue, I'll suggest to revert changes in current file if you agree.

Manvel (changes requested):
Thanks for starting the PR @naarrek 

There are multiple things that you might want to change in here.

As was discussed and mentioned in the comments, let's start solving this issue coming from the other end, first import pages then create logic that uses metadata from those pages, because the decision to take the route you choose might be because you don't understand how you can use them.

## Generic information for importing pages:
- It's a not good idea to use spaces in file and folder names, as those are unsafe characters when used in URLs and being converted into "%20" instead. It's not too bad, I'll suggest to make it a common convention for us, maybe we can adapt from [Google naming](https://developers.google.com/style/filenames) convention, with the exception of special("important" cases) names like "README.md" and "LICENCE". Suggestion, to make it easier to construct and think about those, you maybe can use pathnames from Drupal website, for example: [documentation](https://chrome-automation.com/documentation)/[options](https://chrome-automation.com/options)/[functions-management](https://chrome-automation.com/functions-management), this way you don't need to think too much about the naming. 
- Ensure they are accessible by the same URL as those are in the drupal website, you might use [permalinks](https://cmints.io/en/documentation/pages/frontmatter) for that.
- Ensure to specify the Title of the pages in the front-matter.

## Implementing navigation logic
After importing the page with their metadata you need to decide how you want to create children parent relationship of the pages, there are multiple ways you can do it, either you can create children/parent relationship as it [was described in the issue](https://github.com/browser-automation/cba-website/issues/12#issuecomment-611932147) or by originalPathname of the pages, because that way you might be able to fix the "index" page and permalinks issues mentioned in the review comments. Ideal would be to make the decision from the content creators point of view, there are several pros and cons in each of those decisions, mainly it's the decision of ow to include and exclude items from them appearing in the navigation unless we say that we want all of them to appear there.

Here are several example code snippets you might find useful for each of the solutions
### Using front-matter

See https://github.com/browser-automation/cba-website/issues/12#issuecomment-611932147 to generate the [required data structure](https://github.com/browser-automation/cba-website/blob/master/config.js#L6) and pass the result to `recNav(navigation)`

### Using originalPathname

```js
const headerMenuItems = ["documentation", "tutorial"];
const flyoutPages = site.queryPages(({originalPathname}) => headerMenuItems.includes(originalPathname.split("/")[0]));
const navigation = [];
for (const {originalPathname, pathname, title} of flyoutPages) {
  // TODO: replicate navigation array from config.js using originalPathname.
  // You can use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/split 
  // to split up the originalPathname into parts, create nested structure and populate array.
  // example values of local variables:
  // originalPathname = "documentation/actions/inject";
  // pathname = "inject";
  // title = "Inject";
}
recNav(navigation);
```

pages/Documentation/Actions/bg-function(deprecated).md
```diff
@@ -0,0 +1,10 @@
+---
+title: bg-function(deprecated)
+---
+Because of The CSP rules that google force developers to use in the extension with manifest version 2 there is no possibility to inject code into background page, so that Why I've created new event that calls bg-function.
```
Manvel: Seems like `<br/>`s hasn't been exported and are missing in CSV file.

pages/Documentation/Actions/bg-function(deprecated).md
```diff
@@ -0,0 +1,10 @@
...
+Using bg-function now I can provide ready functions that uses Chrome API.
+Let's look through picture example:
+&lt;$function=removeCookie&gt; 
+&lt;$attr=facebook&gt; 
+as you can see &lt;$function=> place holder specify the function name and &lt;$attr=&gt;  placeholder specify parameters for that function if the function uses several parameters than you need to specify several &lt;$attr=&gt;  placeholders.
+<img src="/sites/default/files/bg-function.jpg" width="500px">
```
Manvel: We also need to import images.

package.json
```diff
@@ -23,7 +23,8 @@
     "url": "https://github.com/Manvel/cmints-single-lang-boilerplate/issues"
   },
   "dependencies": {
-    "cmints": "^0.8.1"
+    "cmints": "^0.8.1",
+    "directory-tree": "^2.2.4"
```
Manvel: Let's remove this if not using.

theme/layouts/_navigation.ejs
```diff
@@ -1,4 +1,32 @@
...
+const regularExpression = /\\|\//;
+const flyoutPages = site.queryPages(({originalPathname}) => headerMenuItems.includes(originalPathname.split(regularExpression)[0]));
+
+let navigation = [];
+let children = [];
+let obj = {};
+let value = {};
+
+for (const {originalPathname, pathname, title} of flyoutPages) {
+  console.log(originalPathname.split('\\')[1], originalPathname.split('\\')[2])
```
Manvel: CMintS is currently [broken on windows](https://github.com/cmints/cmints/issues/195), I think CMintS needs to be first fixed before proceeding with the navigation, as it requires the paths to be correct.

pages/Documentation/Actions/bg-function(deprecated).md
Manvel: Instead is better to use `<p>` in general, or double new line in the markdown.

pages/Documentation/Actions/bg-function(deprecated).md
```diff
@@ -0,0 +1,10 @@
+---
+title: bg-function(deprecated)
+---
```
Manvel: In general it's a good idea to have a new line separattion between front-matter and page content. Wonder if prettier can handle that?

pages/Documentation/Actions/bg-function(deprecated).md
```diff
@@ -0,0 +1,5 @@
...
```
Manvel: Let's avoid special characters in the filenames, meta-information and keep them simple.

Suggestion: `pages/documentation/actions/bg-function.md `

Manvel: Please don't use uppercase in the Directory names. `Documentation` and `Actions`

pages/Documentation/Actions/bg-function(deprecated).md
```diff
@@ -0,0 +1,5 @@
...
+title: bg-function(deprecated)
+---
+
+Because of The CSP rules that google force developers to use in the extension with manifest version 2 there is no possibility to inject code into background page, so that Why I've created new event that calls bg-function. Using bg-function now I can provide ready functions that uses Chrome API. Let's look through picture example: <$function=removeCookie> <$attr=facebook> as you can see <$function=> place holder specify the function name and <$attr=> placeholder specify parameters for that function if the function uses several parameters than you need to specify several <$attr=> placeholders. ![](/sites/default/files/bg-function.jpg)
```
Manvel: Any idea why have all the new lines from the content been removed? 
In the CSV document the content is the following one:
```

"Because of The CSP rules that google force developers to use in the extension with manifest version 2 there is no possibility to inject code into background page, so that Why I've created new event that calls bg-function.
Using bg-function now I can provide ready functions that uses Chrome API.
Let's look through picture example:
&lt;$function=removeCookie&gt; 
&lt;$attr=facebook&gt; 
as you can see &lt;$function=> place holder specify the function name and &lt;$attr=&gt;  placeholder specify parameters for that function if the function uses several parameters than you need to specify several &lt;$attr=&gt;  placeholders.
<img src=""/sites/default/files/bg-function.jpg"" width=""500px"">"
```

Though you should note that In markdown the new lines will be rendered as single line, in order to break the row you need to use `\` or what is better is to construct meaningful paragraphs by double new lines in the markdown file, you can use combination of both to make the content more readable or add `\` instead to all line endings in order to replicate existing Drupal website and leave content development for later on.

Learn more about paragraph and new line in markdown here: https://commonmark.org/help/tutorial/03-paragraphs.html

I'm not adding this comment to file that have this issue, but I see that significant part of imported content are missing the new line characters for some reason. You can compare the CSV and content to find those.

pages/Documentation/Clipboard.md
```diff
@@ -0,0 +1,13 @@
+---
```
Manvel: Please use lowercase file names `clipboard.md` that applies to all other file names that have uppercase.

pages/Documentation/Actions/bg-inject.md
```diff
@@ -0,0 +1,22 @@
+---
+title: bg-inject
+---
+
+Inject script into backroung page of the extension. Abilities:
+
+*   Access to background page of the extension
```
Manvel: Suggestion: Though it's also possible to construct unordered list using `*`, I'll suggest to use `-` instead as, those are more common and intuitive IMO. Also having 1 space after list sign instead of 3 should make the content more readable.

Reference: https://commonmark.org/help/tutorial/06-lists.html

pages/Documentation/Actions/bg-inject.md
```diff
@@ -0,0 +1,22 @@
...
+Tips:
+
+*   Stop project workflow when using asynchronous functions using sendBgInstruction variable "sendBgInstruction = false;"
+*   Don't forget to continue the project workflow after stoping it "sendInstruction();"
+*   use clipboard object for message passing "clipboard\["myKey"\] = value"
+*   Jump to another action "actionToPlay(num);"
+
+Example:
+
+sendBgInstruction = false;  // Stop workflow of the CBA
```
Manvel: This will be rendered as 1 single line, you might want to use code block for "code examples", for that you just need to wrap the content in triple "`" signs.

Reference: https://commonmark.org/help/tutorial/09-code.html

pages/Documentation/Actions/change.md
```diff
@@ -0,0 +1,9 @@
+---
+title: change
+---
+
+Change action mostly are used on textboxes and textareas, they are Triggering focus and change events on DOM element. Change action uses Jquery selector to Change the value of the DOM Element. Also users can use CBA placeholders:
+
+# Placeholders
```
Manvel: Suggestion: Even though in the CSV file `<h1>` is used, making significant accent on content subitem is semantically wrong, let's use `##` instead, to make it more clear eventually that it's `change`'s subitem if we decide to make `change` to be `<h1>`.

pages/Documentation/Actions/click.md
```diff
@@ -0,0 +1,6 @@
+---
+title: click
+---
+
+CBA uses Click action to trigger click event on DOM Element which it finds using Jquery selector.
+(/sites/default/files/Click.jpg =600x)
```
Manvel: Seems like here you have a syntax error?

Reference: https://commonmark.org/help/tutorial/08-images.html

pages/Documentation/Actions/pause.md
```diff
@@ -0,0 +1,5 @@
+---
+title: pause
+---
+
+The action that Pause project workflow, it could be useful if you want to fill some forms manually and then continue with project workflow. When the project workflow reaches that action, play button changes name to "resume" and text under extension icon changes to "||", so the project will wait for your actions to continue workflow, so anytime you can click on stop button and the extension will go back to freeze mode and will wait for your actions. After manual interaction you can click "resume" button and project will continue it's workflow. ![](/sites/default/files/play_part.jpg =100x) ![](/sites/default/files/pause.jpg =600x)
```
Manvel: What is `=600x`? I can see it's used all over the place.

pages/Documentation/Clipboard.md
```diff
@@ -0,0 +1,13 @@
...
+title: Clipboard
+---
+
+"Clipboard" variable is an object that holds your data dAuring chrome process, It's shared object between projects so you can use it even after project playback, there are several ways you can access "Clipboard" object:
+
+*   you can access the object during script injection, please notice if you want to write data to "clipboard" object you have to be sure that you are writing also in main page's content script, as was already told in [injection documentation](inject) inject action injects script also into all iframes of web page, so in case if you would like to write into clipboard object you have to be sure that you are writing also in main page's content script clipboard variable, you can see example of writing into "clipboard" object in image below:
+![](/sites/default/files/clipboard-inject.jpg)
+
```
Manvel: Detail: No space needed, here you are listing list items.

pages/tutorial/video-presentation-for-CBA.md
```diff
@@ -0,0 +1,5 @@
+---
```
Manvel: Detail: `CBA` in the filename is still uppercase, or you can also change to more meaningful filename.

theme/layouts/_navigation.ejs
```diff
@@ -1,9 +1,40 @@
...
+
+for (const {originalPathname, pathname, title} of flyoutPages) {
+orPath0 = originalPathname.split('\\')[0];
+orPath1 = originalPathname.split('\\')[1]
+orPath2 = originalPathname.split('\\')[2]
+
+  if (orPath1 && orPath2) {
+    childSecond = {title, pathname};
+    childFirst = {'title':orPath1, 'pathname':`${orPath0}\\${orPath1}`, 'child':childSecond};
+    obj = {'title':orPath0, 'pathname':orPath0, 'child':childFirst};
```
Manvel: Please use `/` as URL separators instead of `\` and `\\`. That's what should be used.

If you don't want to wait for the CMintS release, you can use:

`npm i git://github.com:cmints/cmints.git#issue_195_1 -D` to get the latest changes with the fixed wrong URL path separators in windows.

theme/layouts/_navigation.ejs
```diff
@@ -1,9 +1,40 @@
...
+        childFirst = {title, pathname};
+        obj = {'title':orPath0, 'pathname':orPath0, 'child':childFirst};
+        navigation.push(obj);
+    } else if (!orPath1 && !orPath2) {
+      obj = {'title':orPath0, 'pathname':orPath0};
+        navigation.push(obj);
+    }
+};
+console.log(navigation)
+%>
```
Manvel: In general you can simplify your task here, looking into your implementation I see that you wish to construct the connection between folder structures using originalPathname which is relative path to the file from `pages` directory(minus index files and file extensions).

The only relevant information for constructing the connection you have here is the top level menu items - `["documentation", "tutorial", "ready-projects", "what's-new"];`.

So your first task is to find out who are the direct children of each of that items, by direct I don't refer to "grand children"(humanly speaking). So let's understand what we understand as direct children, example if you have a structure:
```
documentation
  actions
    inject
    bg-inject
  tutorials
...
```

For documentation direct children are `actions` and `tutorials`, so now you would like to construct a query that would result into those, but not into `inject` or `bg-inject`. As you have this original paths:
```
documentation
documentation/actions
documentation/actions/inject
documentation/actions/bg-inject
documentation/tutorials
```

Now you need to create a query that will get pages for `documentation/actions` and `documentation/tutorials` originalPathname. In order to make that query works you need to ensure that first part of the URL is `documentation` it has "second part" and missing third part. so basically if we were about to use `originalPathname.split("/")`, then we need to ensure that `[0]` is `documentation`, `[1]` exist and `[2]` is missing.

```
for (topLevelItem of topLevelItems) {
  const secondLevel = site.queryPages(({originalPathname}) => originalPathname.startsWith(`${topLevelItem}/`) && originalPathname.split("/").length === 2);
}  
```

or 

```
for (topLevelItem of topLevelItems) {
  const secondLevel = site.queryPages(({originalPathname}) => {
    const items = originalPathname.split("/");
    return (items[0] === topLevelItem && items.length === 2)
  });
}  
```

Then you can construct same logic for 3-rd level items, please note you don't have to hardcode `items.length === 2)` it was for demonstration purposes.

You can also use recursion to nest as much level you need.

package.json
Manvel: This comment is still relevant.

theme/layouts/_navigation.ejs
Manvel: CMintS release is out, so `npm i cmints` should do the trick and fix wrong pathname URLs.

pages/Documentation/Actions/click.md
naarrek: I want to find a way of resizing images. Here `=600x` notation does not work.

pages/Documentation/Actions/pause.md
naarrek: Same here. I want to find a way of resizing images. Here =600x notation does not work.

pages/Documentation/Actions/bg-function(deprecated).md
naarrek: This is done.

package.json
naarrek: Uninstalled already.

pages/Documentation/Actions/bg-function(deprecated).md
naarrek: > a new line separattion between front-matter and page content.

I didn't get the suggestion. Do you mean prettier to test if there is a separation `---` line?

pages/Documentation/Actions/click.md
Manvel: You can use `` tag instead if you want to have fixed width, but I would rather leave fixed with, as image supposedly should take whole available width, which is 640px according to current Drupal website minus paddings. So I'm not sure if there were a reason to use fixed width in Drupal version of the website.

pages/Documentation/Actions/pause.md
Manvel: See https://github.com/browser-automation/cba-website/pull/18#discussion_r425136776 
you don't need fixed width here as well, also the actual width of image above `/sites/default/files/play_part.jpg` is `100px`, so you don't need to write that down specifically.

pages/Documentation/Actions/bg-function(deprecated).md
Manvel: I mean a new line between second `---` separator and actual page content, but I can see that you have addressed this comment in https://github.com/browser-automation/cba-website/pull/18/commits/510fb12b9e3ee673230be539cb0fbb59625e8ddc So this LGTM.

package.json
```diff
@@ -22,10 +22,9 @@
   "bugs": {
     "url": "https://github.com/Manvel/cmints-single-lang-boilerplate/issues"
   },
-  "dependencies": {
-    "cmints": "^0.8.1"
-  },
+  "dependencies": {},
```
Manvel: Let's move CMintS into dependencies as it's not a developer dependency, but actual one required for website generation.

pages/documentation/actions/bg-function.md
```diff
@@ -0,0 +1,11 @@
+---
+title: bg-function(deprecated)
+---
+
+Because of The CSP rules that google force developers to use in the extension with manifest version 2 there is no possibility to inject code into background page, so that Why I've created new event that calls bg-function. Using bg-function now I can provide ready functions that uses Chrome API. Let's look through picture example:
+`
```
Manvel: You can wrap whole expressions into code block, by wrapping in the 3 back-ticks("`"):
```
<$function=removeCookie>
<$attr=facebook>
```

pages/documentation/actions/bg-function.md
```diff
@@ -0,0 +1,11 @@
+---
+title: bg-function(deprecated)
```
Manvel, line 2: Let's use [permalinks](https://cmints.io/en/documentation/pages/frontmatter#permalinks) to make pages accessible by same URL as in Drupal version now. Otherwise we will need to setup permanent redirect(`301`), not to hurt SEO.

pages/documentation/actions/bg-inject.md
```diff
@@ -0,0 +1,23 @@
...
+Tips:
+
+- Stop project workflow when using asynchronous functions using sendBgInstruction variable "sendBgInstruction = false;"
+- Don't forget to continue the project workflow after stoping it `"sendInstruction();"`
+- use clipboard object for message passing `"clipboard\["myKey"\] = value"`
+- Jump to another action "actionToPlay(num);"
+
+Example:
+
+`sendBgInstruction = false;  // Stop workflow of the CBA
```
Manvel: Let's use 3 "`" here, as it's a multiline code https://commonmark.org/help/tutorial/09-code.html

pages/documentation/actions/change.md
```diff
@@ -0,0 +1,15 @@
...
+
+## Placeholders
+
+`_<$unique=>_` - Unique placeholder is generating random number (13 characters length maximum). So in the pic. example the action is used to find Dom Element with the ID of #sometextBox trigger focus event on it change the value to hello concatenated with 3 charecter length random number. 
+
+<img src='/images/Change.jpg' width = '600px'>
+
+`_<$clipboard=>_` - Clipboard placeholder is used to access [clipboard object](clipboard) data. So now you can access any data from clipboard object for change action using current syntax: <$clipboard=propertyName> so this placeholder will call clipboard\["propertyName"\]. 
+
+![](/images/clipboard-placeholder.jpg)
```
Manvel: One of the images have fixed width here, but the other one doesn't, is there a particular reason to have one or another option?

Manvel: Same apply to other images in current PR.

pages/documentation/actions/index.md
```diff
@@ -0,0 +1,16 @@
+---
+title: Actions
+---
+
+Here is the list of CBA Action Types:
+
+- [inject](/documentation/actions/inject)
```
Manvel: Suggestion: We can generate this list dynamically if we use EJS here, I guess.

pages/documentation/actions/change.md
```diff
@@ -0,0 +1,15 @@
+---
+title: change
+---
+
+Change action mostly are used on textboxes and textareas, they are Triggering focus and change events on DOM element. Change action uses Jquery selector to Change the value of the DOM Element. Also users can use CBA placeholders:
+
+## Placeholders
+
+`_<$unique=>_` - Unique placeholder is generating random number (13 characters length maximum). So in the pic. example the action is used to find Dom Element with the ID of #sometextBox trigger focus event on it change the value to hello concatenated with 3 charecter length random number.
```
Manvel: You have here italic code inside code block, I think you want either of both, but not both. I would go with inline code block as it's about code.

pages/documentation/actions/change.md
```diff
@@ -0,0 +1,15 @@
...
+
+Change action mostly are used on textboxes and textareas, they are Triggering focus and change events on DOM element. Change action uses Jquery selector to Change the value of the DOM Element. Also users can use CBA placeholders:
+
+## Placeholders
+
+`_<$unique=>_` - Unique placeholder is generating random number (13 characters length maximum). So in the pic. example the action is used to find Dom Element with the ID of #sometextBox trigger focus event on it change the value to hello concatenated with 3 charecter length random number. 
+
+<img src='/images/Change.jpg' width = '600px'>
+
+`_<$clipboard=>_` - Clipboard placeholder is used to access [clipboard object](clipboard) data. So now you can access any data from clipboard object for change action using current syntax: <$clipboard=propertyName> so this placeholder will call clipboard\["propertyName"\].
```
Manvel: Why are you using back spaces before "[" and "]"? I have noticed that in the other parts of the PR as well.

pages/documentation/actions/change.md
Manvel: Same applies to another code block below.

pages/documentation/actions/change.md
Manvel: General suggestion: You can always wrap code related content in the inline or multiline code blocks when/if applicable.

pages/documentation/actions/submit-click.md
```diff
@@ -0,0 +1,7 @@
+---
+title: submit-click
+---
+
+Click-submit action is simmilar to click action but besides click action the action is telling CBA that it must wait for page's update, so it could be usefull to use it for clicking submit buttons for example.
```
Manvel: Detail: `page's` -> `page's`

pages/documentation/clipboard.md
```diff
@@ -0,0 +1,13 @@
+---
+title: clipboard
+---
+
+"Clipboard" variable is an object that holds your data dAuring chrome process, It's shared object between projects so you can use it even after project playback, there are several ways you can access "Clipboard" object:
+
+- you can access the object during script injection, please notice if you want to write data to "clipboard" object you have to be sure that you are writing also in main page's content script, as was already told in [injection documentation](documentation/actions/inject) inject action injects script also into all iframes of web page, so in case if you would like to write into clipboard object you have to be sure that you are writing also in main page's content script clipboard variable, you can see example of writing into "clipboard" object in image below:\
+![](/images/clipboard-inject.jpg)
+
```
Manvel: Detail: Seems like there is a trailing new line.

pages/documentation/functions/clear-cookies.md
```diff
@@ -0,0 +1,17 @@
+---
```
Manvel: Interesting seems like this page is not accessible in Drupal version -> https://chrome-automation.com/clear-cookies

Let's add `_` in the beginning of the filename, as it will make a draft out of this page and have a look into it later.

pages/index.md
```diff
@@ -0,0 +1,14 @@
+---
+title: Homepage
+---
+
+1. CBA gives ability to automate chrome browser.
```
Manvel, line 5: Suggestion: you can use trick below to avoid numerating each line:
```
1. CBA gives ability to automate chrome browser.
1. ability to setup step-by-step code injections from Extensions popup interface.
1. ...
...
```

theme/layouts/_navigation.ejs
```diff
@@ -1,17 +1,48 @@
-<nav id="main-navigation">
+<%
+const menuItems = ["", "documentation", "tutorial", "ready-projects",
```
Manvel: Suggestion: This list can go into `config.js` that way we simplify content writers job, so they don't need to modify theme code.

theme/layouts/_navigation.ejs
```diff
@@ -1,17 +1,48 @@
...
+    }
+  }
+  return nav;
+}
+
+const navigation = [];
+for (const menuItem of menuItems) {
+  const page = query(({originalPathname}) => originalPathname === menuItem)[0];
+  navigation.push(getSubPages(page));
+}
```
Manvel: As we are populating navigation dynamically, we can now remove the navigation mock from `config.js`

theme/layouts/_navigation.ejs
```diff
@@ -1,17 +1,48 @@
...
+const menuItems = ["", "documentation", "tutorial", "ready-projects", 
+                   "what's-new"];
+const query = site.queryPages;
+
+function getSubPages(nav)
+{
+  const subPages = query(({originalPathname}) => {
+    const parentPath = `${nav.originalPathname}/`;
+    return originalPathname.startsWith(parentPath) &&
+           !(originalPathname.substring(parentPath.length).includes("/"));
```
Manvel, line 9: We still need to figure out how to include or exclude items from the navigation, that way content writers will be able to decide which page goes into menu and which not.

Manvel: For example we can introduce, something like `menu`:
```
---
tile: Change
menu: true
---
```

theme/layouts/_navigation.ejs
```diff
@@ -1,17 +1,48 @@
...
+  });
+  if (subPages.length) {
+    if (!nav.sub_items){
+      nav.sub_items = [];
+      for (const subPage of subPages) {
+        nav.sub_items.push(subPage);
+        getSubPages(subPage);
+      }
```
Manvel: Currently there is no order in which the items appear in the navigation, in order to fix that we can introduce a weight, so we can sort them out, for example:

```
---
title: Change
weight: 5
---
```

Manvel (changes requested):
Good job converting HTML to Markdown, I have left some comments/suggestions to the markdown part and added notes we have discussed here and there to the navigation part.

pages/documentation/actions/change.md
naarrek: I just resize images that are too big by its original size and extends or equals content  sizes. the other ones are just OK. This i suppose we might want to change if setting responsive properties.

Manvel: > I just resize images that are too big by its original size

They wouldn't stretch more than they parent content if you use for example `max-width: 100%` on images in content. So if you leave default behavior of width it will maximum shrink to 640px on desktop and adjust accordingly on mobile, so you don't need to adjust size for smaller screens. I think you want to stop hardcoding images. To test that you can change width of the main block to `640px`.

pages/documentation/actions/index.md
naarrek: ejs code does not work here even when I change file extension to .ejs.

Manvel: Which code doesn't work?

Can you post a snippet so I can try troubleshooting?

naarrek: > Can you post a snippet so I can try troubleshooting?

later at evening.

naarrek: This suggestion is implemented.

pages/documentation/interface/index.ejs
```diff
@@ -0,0 +1,17 @@
...
+<%
+interfaceList = site.queryPages(({originalPathname}) => {
+return originalPathname && originalPathname.split('/')[0] == 'documentation' && 
+    originalPathname.split('/')[1] == 'interface'}) %>
+<ul>
+<% for (const {title, pathname, originalPathname} of interfaceList) { 
+    if (originalPathname !== 'documentation/interface') { %>
+    <li><a href = '<%= pathname %>'><%= title %></a></li>
+    <% }
+} %>
```
Manvel: From what I understood, we are now duplicating this logic on some pages. I think there are ways to avoid duplicating same logic. From top of my head:
- Creating a new EJS helper and passing along with templateData.
- Creating a new type of layout in `theme/layouts` and rendering those pages accordingly.

pages/documentation/actions/index.ejs
```diff
@@ -0,0 +1,16 @@
+---
+title: Actions
+permalink: actions
+---
+Here is the list of CBA Action Types:
+<%
+actionsList = site.queryPages(({originalPathname}) => {
+return originalPathname && originalPathname.split('/')[0] == 'documentation' && 
+    originalPathname.split('/')[1] == 'actions'}) %>
+<ol>
```
Manvel: In the Drupal version and on other pages you were using unordered lists, let's use it here as well. `<ul>`

pages/documentation/actions/bg-inject.md
```diff
@@ -0,0 +1,24 @@
+---
+title: bg-inject
+permalink: bg-inject
+---
+
+Inject script into backroung page of the extension. Abilities:
```
Manvel: Detail: Abilities goes into a new paragraph.

pages/documentation/clipboard.md
```diff
@@ -0,0 +1,12 @@
+---
+title: clipboard
+permalink: clipboard
+---
+
+"Clipboard" variable is an object that holds your data dAuring chrome process, It's shared object between projects so you can use it even after project playback, there are several ways you can access "Clipboard" object:
```
Manvel: Detail: `dAuring ` -> `during `

pages/documentation/options/index.ejs
```diff
@@ -0,0 +1,20 @@
+---
+title: Options
+permalink: options
+---
+
+Section that describe extension options page. Options:
```
Manvel: Detail: two new lines are missing before `Options: `

pages/documentation/options/cba.md
```diff
@@ -0,0 +1,8 @@
+---
+title: CBA
+permalink: cba-big
+---
+
+The option providing bigger CBA interface where user can prepare projects without fear that popup window could be closed. It's actually same interface you have in popup window without project play and record buttons. While the popup can be closed in case you click outside here you can be sure that the interface will not be closed by mistake: 
+
+![](/images/Options.png)
```
Manvel, line 13: Seems like current image is missing.

pages/documentation/options/functions-option.md
```diff
@@ -0,0 +1,12 @@
...
+permalink: functions-management
+---
+
+In this section you can create ready functions which you can use in functions interface - just drag and drop the function to actions table to use already ready script in the project. You can find ready functions in the functions section of the documentation. If you created some cool function I'll ask you to send it to me using mail form, or write to me [directly](mailto:saroyanm@gmail.com?Subject=From%20CBA%20user) and after manuall review of the function it could be possible that the function will appear in functions section and will be available to everyone. To create CBA function you need to:
+
+- Define function name
+- write script in script section
+- and choose action type
+
+![](/images/functions_screen.png)
```
Manvel, line 19: Seems like current image is missing.

pages/documentation/options/functions-option.md
```diff
@@ -0,0 +1,12 @@
+---
+title: Functions options
+permalink: functions-management
+---
+
+In this section you can create ready functions which you can use in functions interface - just drag and drop the function to actions table to use already ready script in the project. You can find ready functions in the functions section of the documentation. If you created some cool function I'll ask you to send it to me using mail form, or write to me [directly](mailto:saroyanm@gmail.com?Subject=From%20CBA%20user) and after manuall review of the function it could be possible that the function will appear in functions section and will be available to everyone. To create CBA function you need to:
```
Manvel: Detail: Let's replace new lines as can be seen in Drupal with paragraphs for now.

pages/documentation/options/import-export.md
```diff
@@ -0,0 +1,8 @@
+---
+title: Import-export
+permalink: importexport
+---
+
+In this section user can export and import already created projects or groups. To export ready project or group just click on corresponding item in export table and copy generated JSON string. To import ready project, just past the JSON string into import TextArea and choose corresponding group you would like to import projects to, or choose root to create new group automatically and click on import button:
```
Manvel: Detail: Let's replace new lines as can be seen in Drupal with paragraphs for now.

pages/tutorial/index.ejs
```diff
@@ -0,0 +1,11 @@
+---
+title: Tutorial
+---
+
+<h1> Tutorials </h1>
+<%
+tutorialList = site.queryPages(({originalPathname}) => {
+return originalPathname && originalPathname.split('/')[0] == 'tutorial'})
+for (const {title, pathname, originalPathname} of tutorialList) { %>
+    <h2> </p><a href = '<%= pathname %>'><%= title %> (video)</a></h2>
```
Manvel: On other pages you were using unordered lists, for listing subitems, let's use same here as well.

pages/tutorial/recording.md
```diff
@@ -0,0 +1,8 @@
+---
+title: Recording
+permalink: video-demo1
+---
+
+CBA store all information in browser Local Storage so that why after closing the chrome browser and restarting the computer records are still available. Also you can record your interaction by clicking the "rec" button. After clicking record button "rec" message appear under CBA icon that mean that you can start to interact with web page. After finishing the interaction just click on "stop" button. Now you can manage and edit your automation.
```
Manvel: The content of this page should be an embeded youtube video.

pages/documentation/actions/bg-inject.md
Manvel: See -> https://github.com/browser-automation/cba-website/pull/18#discussion_r422863492

pages/documentation/options/import-export.md
Manvel: See https://github.com/browser-automation/cba-website/pull/18#discussion_r422863492

pages/documentation/options/functions-option.md
Manvel: See -> https://github.com/browser-automation/cba-website/pull/18#discussion_r422863492

package.json
Manvel: This one is still relevant.

Manvel (changes requested):
Thanks for fixing most of the content related issues.

I have mostly added some small correction and a suggestion about avoiding the duplicated logic on the pages that list sub_pages.

pages/documentation/options/functions-option.md
naarrek: > 
> 
> Detail: Let's replace new lines as can be seen in Drupal with paragraphs for now.

One small question. What is here from readability point of view more preferable "\\" or double new line? I would prefer here "\\".

naarrek: Or maybe adding double new lines everywhere is more preferable?

Manvel: "`\`" Doesn't provide any semantical meaning, as it just breaks the row, but if we want to start from new line(like a paragraph in the book) we rather better use Paragraph(`<p>`). I would vote to use paragraph instead as visually on Drupal website this meant as paragraph and leave rest for later, as I'm anyway planing to rewrite whole documentation.

Also if we won't use `\` in content, content writers can decide to have row char limit without affecting how the sentence will be rendered making MD file more readable, for example -> https://raw.githubusercontent.com/Manvel/blog/master/pages/projects/privacy-manager/rewrite.md.

I will suggest to use paragraphs (double new line) and fix content structure later on when updating the content.

pages/documentation/actions/index.ejs
```diff
@@ -1,16 +1,19 @@
...
-    <li><a href = '<%= pathname %>'><%= title %></a></li>
-    <% } 
-} %>
-</ol>
+    return originalPathname && originalPathname.split('/')[0] == 'documentation' && 
+        originalPathname.split('/')[1] == 'actions'}) %>
+    <ul>
+    <% for (const {title, pathname, originalPathname} of actionsList) {
+        if (originalPathname !== 'documentation/actions') { %>
+            <li><a href = '<%= pathname %>'><%= title %></a></li>
```
naarrek: To be added later: list order by weight.

theme/layouts/_navigation.ejs
```diff
@@ -1,22 +1,54 @@
...
+  if (subPages.length) {
+    if (!nav.sub_items){
+      nav.sub_items = [];
+      for (const subPage of subPages) {
+        if(subPage.menu !== false) {
+        nav.sub_items.push(subPage);
+        getSubPages(subPage);
+        }
+      }
+      nav.sub_items.sort((a, b) => (a.weight > b.weight) ? 1 : -1);
```
Manvel: What if the weight is not defined and what about equal weights? In both cases this will result in negative 
value and items will "go first".

Similar to what we have now with "Interface" subitems, where "Functions datagrid" appear first, while I guess was expected to be last, as in the Drupal version, as weight is `undefined`.

In order to fix that I guess we can use `a.weight - b.weight` that way we will also cover the case of equal items instead and undefined items seem to go to the end if I understand correctly from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort.

theme/layouts/_navigation.ejs
```diff
@@ -1,22 +1,54 @@
...
+    return originalPathname.startsWith(parentPath) &&
+           !(originalPathname.substring(parentPath.length).includes("/"));
+  });
+  if (subPages.length) {
+    if (!nav.sub_items){
+      nav.sub_items = [];
+      for (const subPage of subPages) {
+        if(subPage.menu !== false) {
+        nav.sub_items.push(subPage);
+        getSubPages(subPage);
```
Manvel: Detail: Let's use space separators between `if` and `(subPage.menu !== false)` to make it consistent with other expressions here.

Detail2: indentation is missing in the `if` block.

pages/documentation/actions/bg-function.md
```diff
@@ -0,0 +1,18 @@
...
```
Manvel, line 2: Currently titles are missing from the content pages. I think we can add them in `default.ejs`.

Manvel (changes requested):
Good job @naarrek jan.

Added couple of comments which should be trivial to fix.

Looking forward for the duplication implementation fix and we should be good to go.

theme/layouts/_navigation.ejs
naarrek: > In order to fix that I guess we can use `a.weight - b.weight` 

I tried this. In case of function returning `a.weight - b.weight`  the result goes same in case of interface 'actions datagrid` goes first when expected to go last. Moreover, when one weight is undefined the result is completely unpredictable. In case of same weight the result is unpredictable in local manner (relative only to same weight items).

Manvel: @naarrek This sounds like issue related to `undefined` values, maybe we can try taking care of them as well, maybe?

```
nav.sub_items.sort((a, b) => {
        if (!a.weight)
          return 1;
        if (!b.weight)
          return -1;
        return a.weight - b.weight;
      });
```

Please let me know if that fixes the issue.

config.js
```diff
@@ -3,188 +3,23 @@ const templateData =
...
-            link: ""
-          },
-          {
-            title: "Version 5.0",
-            link: ""
-          }
-        ]
+    menuItems: ["", "documentation", "tutorial", "ready-projects",
+      "what's-new"],
+    pagesList: function (absolutePath) {
```
Manvel: Suggestion: Currently you are hardcoding page's `originalPathname` in EJS files, what you can do instead is pass `pathname` and get `originalPathname` using `site.queryPages `:
```js
function (pathname) {
  const query = templateData.site.queryPages;
  const {originalPathname} = query((page) => page.pathname === pathname)[0];
```

config.js
```diff
@@ -3,188 +3,23 @@ const templateData =
...
+      optionsList = templateData.site.queryPages(({originalPathname}) => {
+        return originalPathname && originalPathname.split('/')[0] == absolutePath.split('/')[0] && 
+            originalPathname.split('/')[1] == absolutePath.split('/')[1]});
```
Manvel: Seems like this logic doesn't work only for direct subpages of the page, but also with subpages of subpages.

```
// Consider we have originalPathname of the current page
query((page) => page.originalPathname.startsWith(originalPathname) &&
  page.originalPathname.replace(originalPathname, "").split("/").length === 2);
```

config.js
```diff
@@ -3,188 +3,23 @@ const templateData =
...
+      arr = [];
```
Manvel: Detail: let's use more meaningful variable name, thinking `subPages`, `pageLists`, `listOfPages`?

Manvel (changes requested):
I have added couple of suggestions, I think we are very close finishing with the review.

config.js
naarrek: > Seems like this logic doesn't work only for direct subpages of the page, but also with subpages of subpages.

don't really get what's the issue... :(

Manvel: Your goal here is to list direct children of `absolutePath` passed above, so for example if you have:
```
/documentation/actions/inject
/documentation/actions/change
/documentation/actions/change/more-change
```

And pass `/documentation/actions` as `absolutePath`, your script will list all the items, while I assume the expected result is to list only:
```
/documentation/actions/inject
/documentation/actions/change
```

There is a fix for that in the comment also.

naarrek: Then if we do like in this -> https://github.com/browser-automation/cba-website/pull/18#discussion_r429625793 comment I think the issue will be resolved.

config.js
```diff
@@ -3,188 +3,35 @@ const templateData =
...
-          }
-        ]
-      }
-    ]
+        if (!a.weight)
+          return 1;
+        if (!b.weight)
+          return -1;
+        return a.weight - b.weight;
+      });
```
Manvel: Suggestion: We are duplicating current function, I think we can reuse it.

We can introduce new method in templateData, maybe ex.: `sortByWeight` and use it both here and in `navigation.ejs`.

config.js
```diff
@@ -3,188 +3,35 @@ const templateData =
...
+
+      const items = pagesQuery.map((item) =>
+        `<li><a href = '${item.pathname}'>${item.title}</a></li>`);
+
+      const result = `<ul>${items.join("")}</ul>`;
+
+      return result;
```
Manvel: Detail: You can just return:
```
return `<ul>${items.join("")}</ul>`;
```

config.js
```diff
@@ -3,188 +3,35 @@ const templateData =
...
-        ]
-      },
+    menuItems: ["", "documentation", "tutorial", "ready-projects",
+      "what's-new"],
+    listOfPages: function (pathname)
+    {
+      const query = templateData.site.queryPages;
+      const {originalPathname} = query((page) => page.pathname === pathname)[0];
+
+      let pagesQuery = query((page) => page.originalPathname.startsWith(originalPathname) &&
```
Manvel: Detail: we are not reassigning current value, so we can use `conts` I guess?

Manvel (approved):
Almost done.

pages/documentation/actions/index.ejs
```diff
@@ -0,0 +1,8 @@
+---
+title: Actions
+permalink: actions
+weight: 2
+---
+Here is the list of CBA Action Types:
+<%- site.listOfPages('actions'); %>
```
Manvel: The idea behind lately introduced changes, to prevent hardcoding current values, you can use here:
```
<%- site.listOfPages(page.pathname); %>
```

Same applies to other index pages, that yet uses hardcoded value

Manvel (changes requested):

Manvel (approved):
Good job @naarrek jan!

PR #19, opened 2020-05-14
Implemented main menu navigation basics

Manvel:
Hey @naarrek jan, as we have discussed, please find the basis of the header menu implementation, it's crude and might need some polishing, but I guess it can be a good start.

Current implementation doesn't contain:
- Sub menu items weight.
- Menu items exclusion/inclusion using front-matter.
- Old data in config.js needs to be cleaned still.


naarrek:
>     * Old data in config.js needs to be cleaned still.
@Manvel should we clean it or delete navigation part??

Issue #9, opened 2020-04-09
Implement sidebar and breadcrumbs


Manvel:
# Background
Sidebar currently is the replication of the header’s main navigation, so you can reuse it here, but besides the sidebar content layout does also have breadcrumbs, which corresponds to the page user is currently viewing. We don’t need to implement the logic of highlighting the breadcrumbs here, but we can at least prepare that component, similar we did for header navigation.
<img width="986" alt="Screen Shot 2020-04-07 at 8 32 10 PM" src="https://user-images.githubusercontent.com/716969/78861768-1154da00-7a4f-11ea-8456-83c7ac5ee200.png">

# What to change
Implement sidebar and breadcrumbs styles.

PR #23, opened 2020-06-01
Issue #9 - Added Breadcrumb

naarrek:


theme/layouts/_breadcrumbs.ejs
```diff
@@ -0,0 +1,16 @@
+<%
+const query = site.queryPages;
+if (page.title !== 'Homepage') {
```
Manvel: We shouldn't rely on title value to filter page, it's better to use in general `pathname`, `originalPath` or add an exclusion using front-matter, as the title can always change.

theme/layouts/_breadcrumbs.ejs
```diff
@@ -0,0 +1,16 @@
...
+const queryPage = query(({originalPathname}) => 
+    originalPathname.split('/').pop() == page.title.toLowerCase().replace(' ','-'));
+const breadcrumbNames = queryPage[0].originalPathname.split('/'); %>
+<div id = breadcrumb>
```
Manvel: This is an invalid attribute declaration.
I think you meant here `id="breadcrumb"`

theme/layouts/_breadcrumbs.ejs
```diff
@@ -0,0 +1,16 @@
+<%
+const query = site.queryPages;
+if (page.title !== 'Homepage') {
+const queryPage = query(({originalPathname}) => 
+    originalPathname.split('/').pop() == page.title.toLowerCase().replace(' ','-'));
```
Manvel: Might be I'm missing something here, but in general It's not a a good idea to construct layout expressions using `page.title` modification. Page title will always change and you don't want to adjust this expression each time that happens.

As this was also case in previous PRs, for example when title was check against constructing sub items. I think it will be a good idea to have some best practices handbook. 

I have created one here https://github.com/browser-automation/cba-website/wiki/Best-practices-handbook, feel free to help me with updating and improving wordings in there, thinking might help as a reference point in future.

theme/layouts/_breadcrumbs.ejs
```diff
@@ -0,0 +1,16 @@
...
+const breadcrumbNames = queryPage[0].originalPathname.split('/'); %>
+<div id = breadcrumb>
+<% for (name of breadcrumbNames) {
```
Manvel: Detail: Unless `const` is used the variable will be global, let's avoid that, this suppose to catch EJS linter, but as we don't have one, we need to be careful here :(

theme/layouts/_breadcrumbs.ejs
Manvel: I think breadcrumbs semantically are either ordered lists or unordered lists. As we are showing the structure, like a path to a page sounds more like ordered one for me.

Example: https://developer.mozilla.org/en-US/docs/Web/CSS/Layout_cookbook/Breadcrumb_Navigation

theme/layouts/_breadcrumbs.ejs
```diff
@@ -0,0 +1,16 @@
...
+    const queryPath = query(({title}) => name == title.toLowerCase().replace(' ','-'));
+    const lastBread = name !== breadcrumbNames[breadcrumbNames.length -1];
+    %>
+    <span <%- lastBread ? "class = 'breadcrumb-items'" : "" %> >
+    <a <%- lastBread ? "href = 'queryPath[0].pathname'" : "" %>><%= queryPath[0].title %></a>
```
Manvel: Links doesn't work, they are linking to `http://127.0.0.1:3000/queryPath[0].pathname` when testing on localhost

Manvel: Interesting, I have just learned the usecase of `<a>` tag without `href` attribute in the breadcrumbs, I was thinking it either an invalid markup, or semantically incorrect, but seems like it is semantically correct [according to the Standards](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-a-element), also from that document:

> If a site uses a consistent navigation toolbar on every page, then the link that would normally link to the page itself could be marked up using an a element:
```
<nav>
 <ul>
  <li> <a href="/">Home</a> </li>
  <li> <a href="/news">News</a> </li>
  <li> <a>Examples</a> </li>
  <li> <a href="/legal">Legal</a> </li>
 </ul>
</nav>
```

I think this is applicable to this case as well, good job!

theme/layouts/_breadcrumbs.ejs
Manvel: Added a relevant note, which might be also helpful in the comment below.

Manvel (changes requested):
Thanks for the PR.

I like that you have implemented a meaningful part and upload the PR without waiting for the Sidebar to be ready, it was an efficient job.

I have added some comments, guess after addressing those the CSS might change, so I'll review that afterwards.

theme/layouts/_breadcrumbs.ejs
naarrek: @Manvel In Drupal website there's no list-structure. What will be better to change markup or follow the recommendations?

theme/layouts/_breadcrumbs.ejs
naarrek: I think this part must be rewritten, because ejs escapes href value, and there's no way to make it not to escape.

theme/layouts/_breadcrumbs.ejs
naarrek: @Manvel  the problem here is that page's permalinks and pathnames are the same. Also `page` variable does not have `originalPathname` attribute. The only option remains the titles I guess...

theme/layouts/_breadcrumbs.ejs
Manvel: As we haven't been following Drupal markup so far, I wouldn't suggest following it now. Following recommendations and best practices, is the best we can do.

theme/layouts/_breadcrumbs.ejs
Manvel: > The only option remains the titles I guess. 

Title is used to define page's title in the `<head>`, see -> https://developer.mozilla.org/en-US/docs/Web/HTML/Element/title for more information. It's up to the content creator to choose the title and developers shouldn't enforce specific titles.  

Please let me know if that informations should go into the [Best-practices-handbook](https://github.com/browser-automation/cba-website/wiki/Best-practices-handbook) or let me know if it's still unclear why we shouldn't use it.

> the problem here is that page's permalinks and pathnames are the same

Sure, they are same. But why is it a problem? I guess I'm missing details here.

Also it shouldn't be much different from using title, I mean if you would do:
```
originalPathname.split('/').pop() == page.pathname;
```

Then you won't need to do all the modification, with changing title into lowercase and do replacements.

But I agree with you that it's not an elegant solution.

> Also page variable does not have `originalPathname` attribute

That's a good point, I don't remember why I didn't include `originalPathname` into the `page`, but I don't mind adding it, if there is a use-case and no argument against.

theme/layouts/_breadcrumbs.ejs
Manvel: No it doesn't escape `"href = 'queryPath[0].pathname'"`. here you are literally returning `href = 'queryPath[0].pathname' ` string, you can try using [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) here.

```
`href="${queryPath[0].pathname}"`
```

theme/layouts/_breadcrumbs.ejs
Manvel: Also if you would like to get current page's originalPath you can do:
```
query({pathName} => page.pathName === pathName)[0].originalPath;
```

theme/layouts/_breadcrumbs.ejs
```diff
@@ -0,0 +1,23 @@
...
+      if (lastBread) { %>
+        <span aria-current='page'><%= queryEachName[0].title %></span>
+      <% }  else { %>
+        <li>
+        <a href = '<%= queryEachName[0].pathname %>'><%= queryEachName[0].title %></a>
+        </li>
+    <% } 
+    } %>
+  </ol>
+</nav>
```
Manvel, line 22: Deatail: In current file you are enclosing strings in both `"` and `'` while in other files it's `"`. Let's keep consistent those.

Manvel: Detail: In current file you are enclosing strings in both " and ' while in other files it's ". Let's keep consistent those.

theme/layouts/_breadcrumbs.ejs
```diff
@@ -0,0 +1,23 @@
+<%
+const query = site.queryPages;
+if (page.pathname !== '') {
+  const queryPage = query(({pathname}) => page.pathname === pathname);
+  const names = queryPage[0].originalPathname.split('/'); %>
```
Manvel: Suggestion: You can use [Destructuring assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) to simplify this expression:
```
const [queryPage] = query(({pathname}) => page.pathname === pathname);
const pathnameParts = queryPage.originalPathname.split('/');
```

theme/layouts/_breadcrumbs.ejs
```diff
@@ -0,0 +1,23 @@
...
+  
+  <nav aria-label="breadcrumb" id = 'breadcrumb'>
```
Manvel: Detail: It's `Breadcrumb`. Also please remove trailing spaces from id part `id="breadcrumb"`

theme/layouts/_breadcrumbs.ejs
```diff
@@ -0,0 +1,23 @@
...
+  <ol>
+    <% for (const name of names) {
+      queryEachName = query(({originalPathname}) => 
+        name == originalPathname.split('/')[originalPathname.split('/').length-1]);
+      lastBread = name == names[names.length -1];
```
Manvel: Suggestion: Thinking, maybe this logic can be simplified. Let's break apart what we are trying to do here.
Let's imagine we are on `documentation/actions/change` page, what I think you are trying to do is to query pages below:
- `documentation`
- `documentation/actions`
- `documentation/actions/change`

So you can just:
```
const lastBreadPath = "documentation/actions/change";
const breadcrumbParts = lastBread.split("/");
for (let i=0; i < breadcrumbParts.length; i++) {
  const originalPathname = breadcrumbParts.slice(0, i).join("/");
  const {pathname, title} = query(...);
}
...
```

Manvel (changes requested):

theme/layouts/_breadcrumbs.ejs
```diff
@@ -0,0 +1,23 @@
...
+  const pathnameParts = queryPage.originalPathname.split('/'); %>
+  
+  <nav aria-label='Breadcrumb' id='breadcrumb'>
+  <ol>
+    <% 
+    for (let i=0; i <= pathnameParts.length; i++) {
+      const pickOriginalPathname = pathnameParts.slice(0, i).join("/");
+      const [{pathname, title}] = query(({originalPathname}) => originalPathname == pickOriginalPathname);
+      if (i == pathnameParts.length) { %>
+        <span aria-current='page'><%= title %></span>
```
Manvel: I guess you meant to use `<li>` instead of span.

theme/layouts/_breadcrumbs.ejs
```diff
@@ -0,0 +1,23 @@
+<%
+const query = site.queryPages;
+if (page.pathname !== '') {
+  const [queryPage] = query(({pathname}) => page.pathname === pathname);
```
Manvel: Detail/Suggestion: This holds all data about page, the query part in the beginning makes it sound like a verb, maybe we can name it just `page`?

theme/layouts/_breadcrumbs.ejs
```diff
@@ -0,0 +1,23 @@
...
+  const pathnameParts = queryPage.originalPathname.split('/'); %>
+  
+  <nav aria-label='Breadcrumb' id='breadcrumb'>
+  <ol>
+    <% 
+    for (let i=0; i <= pathnameParts.length; i++) {
+      const pickOriginalPathname = pathnameParts.slice(0, i).join("/");
+      const [{pathname, title}] = query(({originalPathname}) => originalPathname == pickOriginalPathname);
+      if (i == pathnameParts.length) { %>
```
Manvel: Suggestion: We already have this path part before defining the loop, maybe we can move it outside the loop, to make the code? I think that way we can make code also more readable. What you think?
```
for (let i=0; i < pathnameParts.length; i++) {
  const parentOriginalPathname = pathnameParts.slice(0, i).join("/");
}
<li aria-current="page'"><%= page.title %></li>
```

Manvel (changes requested):
Looks better.

theme/layouts/_breadcrumbs.ejs
naarrek: > 
> 
> Detail/Suggestion: This holds all data about page, the query part in the beginning makes it sound like a verb, maybe we can name it just `page`?

I think `currentPage` is better.

Manvel: Agree

theme/layouts/_breadcrumbs.ejs
naarrek: We need a way to address separate CSS for the current page Breadcrumb part. Also we losing `aria-current='page'`.

Manvel: Why are we losing `aria-current='page' `?

theme/layouts/_breadcrumbs.ejs
```diff
@@ -0,0 +1,21 @@
...
+  const pathnameParts = currentPage.originalPathname.split('/'); %>
+  
+  <nav aria-label='Breadcrumb' id='breadcrumb'>
+  <ol>
+    <% 
+    for (let i=0; i < pathnameParts.length; i++) {
+      const pickOriginalPathname = pathnameParts.slice(0, i).join("/");
+      const [{pathname, title}] = query(({originalPathname}) => 
+        originalPathname == pickOriginalPathname); %>
+        <li class='parts'>
```
Manvel: Suggestion: We don't need to define classname to style this element, instead you can also do `#breadcrumb li:not([aria-current='page'] )`? Which seem to be more descriptive rather than `.parts` and provide cleaner Markup.

theme/less/main.less
```diff
@@ -48,3 +48,34 @@ main
   padding: 15px;
   max-width: 1000px;
 }
+
+#breadcrumb
+{
+  font-size: 0.92em;
```
Manvel: Note: In the Drupal version using `em` does make difference, but on current version changing this to `rem` doesn't affect size, I would suggest to use `rem` instead and adjust sizes accordingly to visually look same comparing the computed size.

You probably know the difference between `em` and `rem`, but I would like to relate to that `rem` is a relative to the HTML element, while `em` to the closest parent. Having dependency on HTML in my experience make the size calculation simpler, rather than when the item is dependent on multiple parents. Sometimes it make sense to use `em` instead `rem`, but in this case I can see that it adds much of value.

theme/less/main.less
```diff
@@ -48,3 +48,34 @@ main
...
+  color: #999;
+  background-color: #f1f1f1;
+  border-radius: 4px;
+  border: 1px solid #e9e9e9;
+  text-shadow: 1px 1px 0px #fff;
+}
+
+#breadcrumb a
+{
+  margin-right: 6px;
```
Manvel: In general consistent left and right margins and/or paddings simplify implementation, especially when it comes to supporting RTL(Right to left) languages it plays huge role. Even though we don't plan supporting RTL at this point, when it's possible to have consistent left and right distances would be great to use that opportunity.

In this case I can see that consistent distances are being created by `#breadcrumb a` and `.parts:after` so instead of this maybe we can just have?
```
.parts:after {
  margin: 0px 6px;
}
```

That way presumably we can just get rid of current rule.

theme/layouts/_breadcrumbs.ejs
```diff
@@ -0,0 +1,21 @@
...
+        <a href = '<%= pathname %>'><%= title %></a>
```
Manvel: Detail: HTML attributes doesn't suppose to have trailing spaces. `<a href='<%= pathname %>'><%= title %></a>`

Manvel (changes requested):
Almost there, please note there is [one old comment](https://github.com/browser-automation/cba-website/pull/23#discussion_r434086964), that is still waiting to be addressed.

Manvel (approved):
Approved, please note there is still one [detail comment](https://github.com/browser-automation/cba-website/pull/23#discussion_r434086964), which hasn't been addressed yet,  but as it's just a detail comment approving right away.

PR #25, opened 2020-06-07
Issue #9 - Sidebar added

naarrek:


config.js
```diff
@@ -27,6 +27,31 @@ const templateData =
         `<li><a href = '${item.pathname}'>${item.title}</a></li>`);
 
       return `<ul>${items.join("")}</ul>`;
+    },
+    getSubPages: function (nav)
```
Manvel: Suggestion: Let's rename `nav` to  something more meaningful, as we are expecting nav to be a page queried by `queryPages`, maybe we can name it just `page` from `queryPages`?

theme/layouts/_sidebar.ejs
```diff
@@ -0,0 +1,35 @@
+<%
+const query = site.queryPages;
+subPages = site.getSubPages;
```
Manvel: Suggestion: As it's a method let's call it for example `getSubPages` in order to make it clear that's not array of subPages. Like already result of calling `site.getSubPages()`

theme/layouts/_sidebar.ejs
```diff
@@ -0,0 +1,35 @@
...
+
+const navigation = [];
+for (const menuItem of site.menuItems) {
+  const page = query(({originalPathname}) => originalPathname === menuItem)[0];
+  navigation.push(subPages(page));
+}
```
Manvel: I assume here you wanted to create the full navigation tree instead of adding just 1 level, maybe:
```
const currentOriginalPath = "documentation/actions/check";
const navigation = [];
for (const menuItem of site.menuItems) {
  const page = query(({originalPathname}) => originalPathname === menuItem)[0];
  if (menuItem === currentOriginalPath.split("/")[0]) {
    page.sub_items = subPages(page);
  }
  navigation.push(page);
}
```

theme/layouts/_sidebar.ejs
```diff
@@ -0,0 +1,35 @@
...
+
+const [currentPage] = query(({pathname}) => page.pathname === pathname);
+let currentPathname = currentPage.originalPathname.split("/");
+pathnameList = [];
+for (let i=currentPathname.length-1; i >= 0; i--) {
+  pathnameList.push(currentPathname.join("/"));
+  currentPathname = currentPathname.slice(0,i);
+}
```
Manvel: Not sure if this is necessary if you will general `navigation` as suggested in the above comment, instead you will only need current page's `originalPathname` in order to not generate children of current page.

theme/layouts/_sidebar.ejs
```diff
@@ -0,0 +1,35 @@
...
+
+  function recNav(navig) { %>
+    <ul>
+      <% for (const {title, pathname, originalPathname, sub_items} of navig) { %>
+        <li <%- originalPathname.split("/").length == 1 ? "class='primary-menu'" : "" %>>
```
Manvel: We don't need to specify `.primary-menu` class to style top most list items to do that you can just use direct children selector `#sidebar ul > li`.

theme/layouts/_sidebar.ejs
```diff
@@ -0,0 +1,35 @@
...
+          <a href="/<%= pathname %>"><%= title %></a>
+            <%
+            for (let i=0; i < pathnameList.length; i++) {
+              if (sub_items && originalPathname == pathnameList[i]) {
+                recNav(sub_items);
+                break;
+              }
+            } %>
```
Manvel: Not sure if this logic is required, here we need to ensure that item has subItems and those subItems are not children of current page.
```
if (sub_items && originalPathname !== currentOriginalPath) {
  recNav(sub_items);
}
```

theme/layouts/_sidebar.ejs
Manvel: Or you can filter out children of current page and reuse existing `recNav` method from `_navigation.ejs`.

theme/layouts/_breadcrumbs.ejs
```diff
@@ -4,7 +4,7 @@ if (page.pathname !== "") {
   const [currentPage] = query(({pathname}) => page.pathname === pathname);
   const pathnameParts = currentPage.originalPathname.split("/"); %>
   
-  <nav aria-label="Breadcrumb" id="breadcrumb">
+  <nav aria-label="Breadcrumb" id="breadcrumb"></nav>
```
Manvel: Detail: I assume change in this file was unintentional.

Manvel (changes requested):
I can see that each time you create a PR it requires less and less comments. Good job.

I haven't reviewed the CSS part, yet, but I'll do it after we finish with the Markup part.

theme/layouts/_sidebar.ejs
naarrek: I think this does not replicate the Drupal website working logic. Not only `currentOriginalPath` matters for viewing children but also we need to specify the logic for "current" parent to view it's sections.

config.js
naarrek: Is it good idea considering that page is reserved by `page.title, page.pathname, ...` varible?

Manvel: Good point, as `page.title` is only accessible in the EJS files I think we are good to go, also if we would have another page variables accessible in the parent scope of the function using local `page` would be fine in this case as you would refer to the local variable not the one specified in the upper scope. So I think we are safe to use `page` here, but feel free to use other meaningful variable if you think of one.

Manvel: Suggestion: We can also introduce second parameter `level`. To help with sidebar implementation, something like:
```
getSubPages: function (nav, level = 100) {
  ...
  if (subPages.length && --level)
  ...
}
```

This way we should be able to get only children of the existing page, which can be useful in sidebar, ex.:
```
site.getSubPages(otherPage, 1);
```

Manvel: Or maybe:
```
getSubPages: function (nav, grandChildren) {
...
  if (subPages.length && grandChildrenl)
...
    templateData.site.getSubPages(subPage, grandChildren);
```

theme/less/main.less
```diff
@@ -42,49 +56,35 @@ header h1 a:hover
...
-{
-  display: flex;
-  list-style: none;
-  padding: 5px 5px 5px 10px;
-  margin: 0;
-}
-
+  body > div
+  {
+    padding: 20px 30px;
```
Manvel: Detail: Seems like we are already specifying this style outside of `@media` query, I guess it's safe to remove it.

theme/less/_sidebar.less
```diff
@@ -0,0 +1,65 @@
+#sidebar
+{
+  font-weight: bold;
+  font-size: 14px;
```
Manvel, line 4: I did notice that we are using precise pixels in some places and relative em/rem in other places, thinking we might need to adjust in future, to make sizing relative which might help with specifying variables.

I have created this -> https://github.com/browser-automation/cba-website/issues/28 

I guess we will tackle that at one point, but it shouldn't be blocker for this issue I guess.

Manvel (approved):
Good job Narek jan. There is just 1 detail comment, which will be great to address before merging, other than that this is ready.

Issue #8, opened 2020-04-09
Implement index page


Manvel:
# Background

[Index page](https://chrome-automation.com/) is the main page user lands and page that has most of the traffic. It’s quite important to optimize the page for the leads, which we eventually will tackle in the future, but for now let’s have a visual representation moved from the Drupal to CMintS.
<img width="967" alt="Screen Shot 2020-04-07 at 8 31 08 PM" src="https://user-images.githubusercontent.com/716969/78861705-e66a8600-7a4e-11ea-8591-76f8c6c84d46.png">

# What to change

Implement index page of CBA website.

PR #29, opened 2020-06-24
Issue #8 - homepage

naarrek:


theme/less/main.less
```diff
@@ -2,6 +2,7 @@
 @import "_navigation.less";
 @import "_sidebar.less";
 @import "_breadcrumb.less";
```
Manvel: We don't need those 3 styles in the homepage, so what about reorganizing our styles so we can only include those for default layout?

Manvel: For that we might also need to re-organize existing styles in main.less to separate file(s).

theme/less/main.less
```diff
@@ -85,5 +86,6 @@ footer
     "sidebar sidebar"
     "main main"
     "footer footer";
+    padding: 20px 120px;
```
Manvel: Isn't padding 120px on mobile screens with limited space too much?

theme/less/_homepage.less
```diff
@@ -0,0 +1,59 @@
...
+    "header header"
+    "main main"
+    "footer footer";
+}
+
+.homepage main article
+{
+  display: grid;
+  grid-template-columns: repeat(auto-fill, minmax(346px, 1fr));
+}
```
Manvel: Why are you using grid for lay-outing homepage content?

I wonder if this just leftover style?

theme/less/_homepage.less
```diff
@@ -0,0 +1,59 @@
...
+
+.box
+{
+  background-color: #f2f2f2;
+  border-radius: 8px;
+  border: 1px solid #dcdcdc;
+  padding: 10px 10px 20px 10px;
+  margin: 20px 20px 0 0;
```
Manvel: Note: your right margin applies to all boxes.

theme/less/_homepage.less
```diff
@@ -0,0 +1,59 @@
...
+}
+
+.box a:active
+{
+  box-shadow: inset 0 0 8px 4px #d4d4d4;
+}
+
+.box a:visited
+{
+  color: inherit;
```
Manvel: I think you want to inherit other anchor related states, like hover, active and also default color, as by default anchor is blue-ish(depends on browsers).

theme/less/_homepage.less
Manvel: If you need it for boxes to have same height, thinking it's should be possible with the flexbox as well.

pages/index.md
```diff
@@ -1,7 +1,10 @@
 ---
+layout: homepage
 title: Homepage
 ---
 
+![](/images/Manvel_440x280_border.jpg)
```
Manvel: This is being wrapped in the `<p>` tag and styled in CSS, I assume it's unintentional, because it's markdown's behavior rather and here you have mix of HTML and Markdown. I wouldn't suggest layouting markdown's behavior. For that I would rather suggest to use EJS or HTML instead for this page.

Manvel: Heading is missing. Guess you meant to reuse the title in the layout file.

pages/index.md
```diff
@@ -1,7 +1,10 @@
...
+
 1. CBA gives ability to automate chrome browser.
 1. Ability to setup step-by-step code injections from Extensions popup interface.
 1. Ability to Record interactions in the chrome browser and playing them.
```
Manvel: Thinking that you might want to wrap content above "boxes", that way it will be simpler for you to separate semantic parts and style. Both wrapping in `<article>` or `<section>` will work. If you use `<section>`.

What you think which wrapper would work better here? 

I'm thinking: if we use `<article>`, that will be consistent semantically with other pages, as we have here the main content of the page, but on the other hand, if we decide in future add more content into the hompage we might want to change that and using `<section>` in that case might be better, choice?

Bit hesitant, let me know what you think about this?

pages/index.md
```diff
@@ -11,3 +14,15 @@ title: Homepage
 1. Ability of using CBA ready functions.
 1. Ability to write your own code that will have access to web page's DOM.
 1. And much much more...
+
+<section class="box">
+  <h3>Install</h3>
+  <p><img src="/images/arrow.png">Get chromium browser automation from google web store.</p>
+  <a href="https://chrome.google.com/webstore/detail/jmbmjnojfkcohdpkpjmeeijckfbebbon?utm_source=chrome-ntp-icon" target="_blank">CBA in chrome web store</a>
```
Manvel: Let's remove unused query string from the url: `https://chrome.google.com/webstore/detail/jmbmjnojfkcohdpkpjmeeijckfbebbon` looks better.

theme/less/_homepage.less
```diff
@@ -0,0 +1,59 @@
...
+  background-color: #f2f2f2;
+  border-radius: 8px;
+  border: 1px solid #dcdcdc;
+  padding: 10px 10px 20px 10px;
+  margin: 20px 20px 0 0;
+}
+
+.box img 
+{
+  float: left;
```
Manvel: Thinking: Using float make sense when img is inside of `<p>` tag, no need to use `` for layouting, it's no more ment to that, developers were using it for layout, when there were no other option.

theme/less/_homepage.less
```diff
@@ -0,0 +1,59 @@
...
+
+.box h3 
+{
+  margin: 0;
+}
+
+.box a
+{
+  text-decoration: none;
+  float: right;
```
Manvel: No need to use float here for layouting button, also note that currently buttons have different distances from each other depending on the size of amount of text inside of `.box > p`. In order to have both buttons on same line, maybe you can use absolute positioning instead and have a specific distance from the bottom of the `.box`.

theme/layouts/homepage.ejs
```diff
@@ -0,0 +1,23 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <% include _head %>
+  </head>
+  <body>
+    <div class="homepage">
```
Manvel: Detail/suggestion: usually I use classes describing page on body, thinking it makes the markup more readable.

Manvel: Also if `.homepage` class is only difference between both layouts, thinking we can find a better way to reuse them. But let's leave that to the end, as the layout might change while addressing comments.

theme/layouts/_breadcrumbs.ejs
```diff
@@ -1,6 +1,5 @@
 <%
 const query = site.queryPages;
-if (page.pathname !== "") {
```
Manvel, line 3: Those changes are related to breadcrumb, I wonder if we need those?

Manvel (changes requested):
Thanks for the PR @naarrek jan.

First round of review comments are ready.

theme/layouts/_breadcrumbs.ejs
naarrek: It's just in order to simplify the code. Do you think it's irrelevant in scope of this PR and should I revert it?

theme/less/_homepage.less
naarrek: I used this in order to be simple with widths and heights of items.

theme/less/_homepage.less
naarrek: I want it to inherit the body default text color.

pages/index.md
naarrek: Now we are using sections inside article. I think can continue going this way, by wrapping all 4 parts in sections. Although in practice I noticed that it is vice versa. They use articles within sections.

pages/index.md
naarrek: During the Prettier test it says that there must be a line. So  I entered a line. Yes, I think it's better to use HTML here.

naarrek: Do we need a heading in homepage?

theme/layouts/_breadcrumbs.ejs
Manvel: Ach right, I forgot that we had a special case for Homepage, LGTM.

pages/index.md
Manvel: Yep: 
Screen Shot 2020-06-25 at 5 42 24 PM

theme/less/_homepage.less
Manvel: Hmm, having a closer look, this might be not a bad idea, only thing is that we might rethink the implementation after adding heading into the `<main>` block.

theme/less/_homepage.less
Manvel: My bad: thought img is outside of the <p>

pages/index.md
naarrek: This will be not usable in case of grid as it requires all items not be wrapped for positioning on homepage.

theme/layouts/homepage.ejs
naarrek: > 
> 
> Detail/suggestion: usually I use classes describing page on body, thinking it makes the markup more readable.

Is this comment still relevant?

Manvel: Yep, because you don't want to call div homepage, but rather say that whole document is homepage.

theme/less/main.less
naarrek: > 
> 
> We don't need those 3 styles in the homepage, so what about reorganizing our styles so we can only include those for default layout?

I think we still need _navigation.less on homepage.

theme/less/_homepage.less
```diff
@@ -1,23 +1,33 @@
...
-.homepage main article
+.homepage div main article
 {
   display: grid;
-  grid-template-columns: repeat(auto-fill, minmax(346px, 1fr));
+}
+
+.homepage div main article h2
+{
+  grid-column: 1 e("/") 3;
```
naarrek: This modification solves the heading problem also takes off risks related to exact sizing in case of `repeat()`.

theme/less/_homepage.less
```diff
@@ -53,7 +66,25 @@
   box-shadow: inset 0 0 8px 4px #d4d4d4;
 }
 
-.box a:visited
+.box a:hover, .box a:active, .box a:visited
 {
   color: inherit;
 }
+
+@media screen and (max-width: 770px)
```
naarrek: I added media query because `grid-column`  doesn't provide automatic features for column number rearranging like it was in case of `repeat()` so we need `grid-column: 1` value here.

pages/index.md
```diff
@@ -1,13 +1,36 @@
 ---
+layout: homepage
```
Manvel: To treat this page as HTML you also need to change the extension of the file to ".HTML" as well.

theme/less/_homepage.less
Manvel: Suggestion: you can add selector `.box a` here as well and remove color style from it above.

theme/less/_homepage.less
```diff
@@ -0,0 +1,90 @@
...
+{
+  position: relative;
+  background-color: #f2f2f2;
+  border-radius: 8px;
+  border: 1px solid #dcdcdc;
+  padding: 10px 10px 60px 10px;
+  margin: 20px 0 0 0;
+}
+
+.box.first
```
Manvel: You probably can use [first-of-type](https://developer.mozilla.org/en-US/docs/Web/CSS/:first-of-type) or [first-child](https://developer.mozilla.org/en-US/docs/Web/CSS/:first-child) instead of adding specific classname.

theme/layouts/_head.ejs
```diff
@@ -18,7 +18,11 @@
 
 <!-- Resources -->
 <script src="<%= site.root %>/js/main.js" defer></script>
-<link href="<%= site.root %>/css/main.css" rel="stylesheet">
+<% if (page.pathname == "") { %>
```
Manvel: Suggestion: As we are applying Layout related style and it's just in our specific case that lyout happens to be index page, what about using:
```
page.layout == "homepage"
```

Instead?

theme/less/_homepage.less
Manvel: What is `e`? I couldn't find anything about e in https://developer.mozilla.org/en-US/docs/Web/CSS/grid-column.

Can you please describe to me this implementation please?

Also I don't understand how this style affects siblings of this selector, I can see that this act as merge for top column of 2 column grid, but I don't understand how and where are we specifying that the grid is 2 column?

pages/index.md
```diff
@@ -1,13 +1,36 @@
...
-1. Ability to Record interactions in the chrome browser and playing them.
-1. Projects saving in browser's local storage.
-1. Step by step code injection.
-1. Ability of projects import and export.
-1. Ability of using CBA ready functions.
-1. Ability to write your own code that will have access to web page's DOM.
-1. And much much more...
+<article>
+
+<h2 class="first">CBA is an extension for chrome browser automation</h2>
```
Manvel: Why are using class `.first` here?

Manvel (changes requested):
Done Narek jan, I have question about the grid implementation, as it's a bit of a mystery for me. As soon I understand the implementation I may provide more meaningful comments.

theme/less/_homepage.less
naarrek: > 
> 
> What is `e`? I couldn't find anything about e in https://developer.mozilla.org/en-US/docs/Web/CSS/grid-column.
> 

It escapes the "/" sign as LESS processes it as "divider" sign.

naarrek: > Also I don't understand how this style affects siblings of this selector, I can see that this act as merge for top column of 2 column grid, but I don't understand how and where are we specifying that the grid is 2 column?

Here 1 / 3 means starting from 1st ending to 3rd column (3rd column does not count). In case of 1 / 1 it's invalid value and it resets to it's default which is to span only the first column. 
As for `grid-column: n` value it sets item to n-th column where n is column number.

naarrek: Here you can play around with this this property. 
https://codepen.io/pen/?&editable=true=https%3A%2F%2Fdeveloper.mozilla.org%2Fen-US%2Fdocs%2FWeb%2FCSS%2FCSS_Grid_Layout%2FAuto-placement_in_CSS_Grid_Layout

Also here are more examples.
https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Grid_Layout/Auto-placement_in_CSS_Grid_Layout

Manvel: > It escapes the "/" sign as LESS processes it as "divider" sign.

Ohh I see, thanks for explanation, sounds like LESS is stuck in his his old implementation. It's sad that we need to escape :(

> Here you can play around with this this property.

This codepen is empty.

> Here 1 / 3 means starting from 1st ending to 3rd column

This yet sounds contr-intuituve to me that specifying `grid-column` on wrapper's first child, sets parent element to 2 column grid, this sounds bit too magical for me, I think they are imitating bootstrap, when you also don't specify parent to grid explicitly AFAIK. Also I couldn't find an example where grid structure is not defined on the wrapper on MDN docs. I'm also bit hesitant, for example if we add 3-rd box below we have to adjust first child and track the grid amount on the children properties rather than on the wrapper which will make things more understandable, for person not familiar with this?

I think if we develop a grid model on CSS we may have more self explanatory implementation, I'm thinking about something like:
```
// 2 stands for 2 column grid on mid sizes?
.grid-md-2 {
  display: grid;
  grid-template-columns: auto auto; // or: 1fr 1fr?
}

// md stands for merge 2 columns on medium size, similar to https://getbootstrap.com/docs/4.0/layout/grid/ 
.col-md-2 {
    grid-column: 1 e("/") 3;
}
```

I think this way we can reuse implementation also on the main layouting and introduce 3 or 4 column grid when/if needed. As we only need 2 column grid I think it's fine, this implementaion has it's own cons, for example if we decide to introduce more than 2 boxes, let's say 4, then we need to create 4 column grid and adjust parents which we could have avoided by wrapper, but I think for now this is fine, unless we feel the need to restructure.

Also current implementation will allow us to be more flexible regarding the content of the homepage, ex.: if we use `<h1>` instead of `<h2>` we don't need to adjust styles, or if we use `<ul>` instead of `<ol>`.

Please let me know what you think @naarrek especially I wonder if it's only me for whom not specifying the grid model on parent item feels too magical. Also feel free to find different naming or maybe you can think of a better implementation, as you are more advanced than me on this already :)

Manvel: Also I wonder if we need `md`, considering we have only 1 size. Also hesitant about that, will wait on your thoughts though.

pages/index.html
```diff
@@ -0,0 +1,66 @@
...
+    <h3>Install</h3>
+    <p>
+      <img src="/images/arrow.png" />
+      Get chromium browser automation from google web store.
+    </p>
+    <a
+      href="https://chrome.google.com/webstore/detail/jmbmjnojfkcohdpkpjmeeijckfbebbon"
+      target="_blank"
+      >CBA in chrome web store</a
+    >
```
naarrek: This kind of disorder is forced by prettier. Can we change the corresponding rule?

Manvel: I think it's fine, because href is quite long and in that case it moves each attribute to it's line. Probably there is a way, but having all on 1 line is also a not a good idea.

pages/index.html
```diff
@@ -0,0 +1,66 @@
+---
+layout: homepage
+title: Homepage
+---
+
+<article>
+  <h2 class="first-grid-item">
+    CBA is an extension for chrome browser automation
+  </h2>
```
Manvel: You can move this into homepage.ejs and use `page.title`, as the value.

naarrek: If we change `page.title`'s value it will change the Homepage button title as well on navigation menu. Do you want to change it to "CBA is an extension for chrome browser automation"?

Manvel: You can introduce: a new front-matter property ex.: `menu`. so `title` will work as a fallback for menu items that doesn't have that prop.

Manvel: THis way we will be consistent with Druapal, as if you notice in Drupal version, the title of the page is `CBA is an extension for chrome browser automation ` instead of `homepage`.

naarrek: > 
> 
> You can move this into homepage.ejs and use `page.title`, as the value.

Ok, but did you mean outside of `article` wrapper? If so there would be no need for `grid-column` property for first-child anymore, because it is no child of `article` wrapper anymore.

Manvel: From [MDN ](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/article) : 

> `Each <article> should be identified, typically by including a heading (<h1>-<h6> element) as a child of the <article> element`

Title should be located inside of `<article>`, so you might want to move `<article>` as well, that way you also will be consistent with the `default.ejs`(I guess).

pages/index.html
```diff
@@ -0,0 +1,59 @@
+---
+layout: homepage
+title: CBA is an extension for chrome browser automation
+menu: Homepage
+---
+
+<img src="/images/Manvel_440x280_border.jpg" alt="CBA-banner" />
```
Manvel: Detail: You can use spaces in alt attribute. `CBA banner `.

pages/index.html
```diff
@@ -0,0 +1,59 @@
...
+</section>
+
+<section class="box">
+  <h3>Documentation</h3>
+  <p>
+    <img src="/images/doc_grey50.png" />
+    Learn more about how CBA works.
+  </p>
+  <a href="/documentation">Documentation</a>
+</section>
```
Manvel: Even though grid solves quite a lot keeping HTML semantically nice and good looking. I'm thinking that this is quite fragile and error prone implementation for a page from the non developer point of view who don't know the context of the decisions and css implementation details.

I can't think of any solution right now from top of my head that will be ideal for this case.

Thinking, let's leave implementation this way for now until a better suggestion/sollution  is made.

theme/less/_homepage.less
```diff
@@ -0,0 +1,95 @@
+.homepage div
```
Manvel: As you only load this css in homepage, guess you don't need to specify `.homepage` class anymore.

theme/less/_homepage.less
```diff
@@ -0,0 +1,95 @@
...
+    "footer footer";
+}
+
+.homepage main article
+{
+  display: grid;
+  grid-template-columns: auto auto;
+}
+
+.homepage main article > :first-child
```
Manvel: Detail: You don't supposedly need `>` here. `.homepage main article:first-child`

theme/less/_homepage.less
```diff
@@ -0,0 +1,95 @@
...
+
+  .homepage main article > img
+  {
+    margin: 0 auto;
+  }
+
+  .box:first-of-type
+  {
+    margin: 20px 0 0 0;
+  }
```
Manvel: Question: Why we want exactly the first `.box` item to have `margin-top` ?

theme/less/_homepage.less
```diff
@@ -0,0 +1,95 @@
...
+
+  .homepage main article > :first-child
+  {
+    grid-column: 1;
+  }
+
+  .homepage main article > img
+  {
+    margin: 0 auto;
+  }
```
Manvel: Question: Why are we centering image here?

theme/less/_homepage.less
```diff
@@ -0,0 +1,95 @@
...
+{
+  .homepage main article
+  {
+    grid-template-columns: auto;
+  }
+
+  .homepage main article > :first-child
+  {
+    grid-column: 1;
+  }
```
Manvel: Thoughts: I still thinking that maintaining such specific implementation feels wrong for content.

Might be that it's me who used to declarative way of grid implementation... But I sense something off about it, can't really say/define what right now.

theme/less/_main.less
```diff
@@ -0,0 +1,85 @@
+body
```
Manvel: I think main is quite ambiguous name for this file, maybe if we separate the content of this file, we can have a more meaningful name or main will fit more?

See suggestions below:

theme/less/_main.less
```diff
@@ -0,0 +1,85 @@
...
+header a
+{
+  text-decoration: none;
+  color: #262626;
+}
+
+header h1 a:hover
+{
+  color:#3253ad;
+}
```
Manvel: Suggestion/separation: What about moving this into `_header.less`?

theme/less/_main.less
```diff
@@ -0,0 +1,85 @@
...
+
+footer
+{
+  grid-area: footer;
+  border-top: 1px solid #d6d6d6;
+  padding: 40px;
+  line-height: 15px;
+  text-align: center;
+  font-size: 15px;
+}
```
Manvel: Suggestion/separation: What about moving this into `_footer.less`?

theme/less/_main.less
```diff
@@ -0,0 +1,85 @@
...
+main
+{ 
+  grid-area: main;
+  padding: 15px 0;
+}
+
+img
+{
+  max-width: 100%;
+}
```
Manvel: Suggestion/separation: What about moving this into `_content.less`?

Also changing selector to: `main img` or `main article img` maybe? As we want to use this style only inside of the content.

Manvel: Also maybe we can move `main` selector above to `_content.less` As main seem to be wrapper of our "main" content and relevant to it indirectl6.

theme/layouts/homepage.ejs
```diff
@@ -0,0 +1,32 @@
...
+        <h1>
+          <a href="\">
+            <%= site.title %>
+          </a>
+        </h1>
+        <p>
+          <%= site.subtitle %>
+        </p>
+        <% include navigation/_main %>
+      </header>
```
Manvel: Suggestion/separation: What about moving this into `_header.ejs`, this way we will have less duplications in our layouts and it will be consistent with other separations we had.

theme/layouts/navigation/_main.ejs
```diff
@@ -15,8 +15,8 @@ for (const menuItem of site.menuItems) {
 
 <nav id="mobile-nav">
 <select>
-   <% for ({title, pathname, sub_items} of navigationPages) { %>
-      <option value="<%- pathname %>"><%- title %></option>
+   <% for ({title, menu, pathname, sub_items} of navigationPages) { %>
+      <option value="<%- pathname %>"><%- menu ? menu : title %></option>
```
Manvel, line 19: Guess we also need to add such solution to the "breadcrumbs"?

theme/less/_homepage.less
```diff
@@ -0,0 +1,95 @@
...
+}
+
+.homepage main article > :first-child
+{
+  grid-column: 1 e("/") 3;
+}
+
+.box
+{
+  position: relative;
```
Manvel: As we are now introducing possitioning we might need a `z-index` for our header menu items, as this currently is overflowing the top menu absolute positioning.

Manvel (changes requested):
It was harder to review this than other PRs, as I have mixed feelings about homepage implementation - semantically it feels nice, but yet quite fragile and somehow off, I'm having hard time to define it, but it might that I'm biased to the classical grid implementation, when you have classes and wrappers are self explanatory because of their declarative nature.

Anyway thinking it's better to leave it this way, maybe I'm just overthinking.

pages/index.html
naarrek: I was thinking about flex implementation, but there must be fixed width of the row for flex-wrap property to function which brings to difficulties with calculations of actual width. That would be more obvious way of styling but again it will depend on max-width value thus will be risky. I think of current implementation is not very obvious, but it's easy to manipulate when it comes to adding new grid items. `grid-template-columns` and `grid-column` can in specific cases interfere with each other but it will be easy to take care of conflict considering that that `grid-column` (or `grid-column-start` and `grid-column-end`) is for positioning items in grid, `grid-template-columns` is for defining grid template. Anyway, I understand your concerns about fragility (because the properties can interfere with each other, but not at current values) but like this kind of implementation for it's semantic simpleness.

theme/less/_homepage.less
naarrek: We have a problem here. When I removed the `.homepage` the page shows grid configurations of `_main.less`  (particularly `body > div`) because we import `_main.less`  in `homepage.less`. We need to separate those  configurations in other file too or to move it to `main.less`. Also we can have one `homepage.less` unified with `_homepage.less`. And of course have a `_generic.less` for general properties.

theme/less/_homepage.less
naarrek: This will search for article elements and select first of them. The space before pseudo-element is mandatory here. And the `>` sign I entered for selector not to select other element's first-childs like `section h3` (although there's no other grids inside `article` so this -> `grid-column` particular property I think will not take effect on that childs).

theme/less/_homepage.less
naarrek: The difference with other views is not `margin-top` but `margin-right: 0`.

theme/less/_homepage.less
naarrek: I don't know it seemed to me more esthetic. It's not principal.

pages/index.html
Manvel: > but there must be fixed width of the row for flex-wrap property 

Not sure if I understand this, but whole point flexbox is that you don't need fixed width.

> but it's easy to manipulate when it comes to adding new grid items

Easy for you, but not for content writers. Either you provide content writers with obvious layout management classes, or limit content they can touch.

I was more thinking that we might be able to move most of the implementation details out of the `page/index.html`. Maybe:
```
index.md
---
title: Homepage
menu: Homepage
layout: homepage
hero-image: /images/Manvel_440x280_border.jpg
box-download-title: Download
box-download-content: Get chromium browser automation from google web store.
box-download-button: CBA in chrome webstore
...
---

1. CBA gives ability to automate chrome browser.
1. Ability to setup step-by-step code injections from Extensions popup interface.
1. Ability to Record interactions in the chrome browser and playing them.
...
```

But even for this implementation text content needs to be wrapped and all the effort to avoid them will gone.

Or I was thinking that maybe we can provide some self explanatory styling options, so the whole page will be less error prone to edit.

theme/less/_homepage.less
Manvel: > When I removed the .homepage the page shows grid configurations of _main.less (particularly body > div) because we import _main.less in homepage.less

What I'm trying to say is that you don't have to introduce another class to change those predefined values, so `body > div ` can also work, or loading default layout styles only for default layout. Maybe after you separate other style mentioned below best solution might be easier to choose.

naarrek: What about `_content.less`?

theme/less/_homepage.less
Manvel: Strange that it works, because space doesn't mean anything AFAIK. Then you probably want here is  `.homepage main article > *:first-child` `*` stands for any element.

theme/less/_homepage.less
Manvel: > The difference with other views is not margin-top but margin-right: 0.

What about avoiding using `margin-right: ` for first box on the first place?

theme/less/_homepage.less
Manvel: > _content.less

What about it? Not sure if I'm following.

theme/less/_homepage.less
Manvel: Solutions from top of my head(there are probably more):
- Using equal left and right margins for both boxes.
- Using https://developer.mozilla.org/en-US/docs/Web/CSS/column-gap

theme/less/_homepage.less
```diff
@@ -0,0 +1,82 @@
...
+  grid-template-columns: auto auto;
+  column-gap: 20px;
+}
+
+main article > *:first-child
+{
+  grid-column: 1 e("/") 3;
+}
+
+.box
```
naarrek: Can not figure out why widths of boxes are different. You can check out chrome DevTools size measurement tool.

public/css/homepage.css
```diff
@@ -0,0 +1,190 @@
+header {
```
Manvel: We can add generated files to the `.gitignore`.

theme/less/_homepage.less
naarrek: Changing `grid-template-columns: auto auto;` to `grid-template-columns: 1fr 1fr;` adjusts the sizes.

Manvel: Good job resolving this.

theme/less/_homepage.less
```diff
@@ -0,0 +1,82 @@
...
+}
+
+@media screen and (max-width: 770px) 
+{
+  main article
+  {
+    grid-template-columns: auto;
+  }
+
+  main article > :first-child
```
Manvel: Detail: This selector is inconsistent with the parent one and still thinking maybe we can use a declarative naming here, as it's about page's content manipulation. `.merge-cells`?

theme/less/_homepage.less
```diff
@@ -0,0 +1,82 @@
+body div
```
Manvel: Let's move content into `hompage.less`.

theme/less/_content.less
```diff
@@ -0,0 +1,26 @@
...
+{
+  max-width: 960px;
+  margin: 0 auto;
+  padding: 20px 30px;
+}
+
+main article img
+{
+  max-width: 100%;
+}
```
Manvel, line 4: Let's move all beside current style into `body-wrapper.less`. As agreed.

Manvel (changes requested):
Almost there!

theme/less/_homepage.less
naarrek: I don't think that here we exactly are merging cell, because the property `grid-column: 1` (or grid-column: 1 / 2) means there's only one column.

public/css/homepage.css
```diff
@@ -1,3 +1,18 @@
+body {
```
naarrek: homepage.css is added to `.gitignore` but that does not take effect. I think it's windows-ic issue.

Manvel: I guess you also need to delete file and make another commit with deleted file, as it's being tracked since it was added first time.

.gitignore
```diff
@@ -1,5 +1,7 @@
 node_modules
 public/css/main.css
+public/css/content.css
```
Manvel: We don't generate content.css, so we should be save skipping it.

theme/less/_homepage.less
Manvel: Okey, agree leaving as it is right now.

pages/index.html
Manvel: To be followed up in -> https://github.com/browser-automation/cba-website/issues/30

theme/less/_body-wrapper.less
```diff
@@ -0,0 +1,21 @@
...
+  max-width: 960px;
+  margin: 0 auto;
+  padding: 20px 30px;
+}
+
+@media screen and (max-width: 770px)
+{
+  body > div
+  {
+    padding: 20px 50px;
```
Manvel: Suggestion: on mobile left and right paddings are precious, let's change this to at least 20px.

theme/layouts/homepage.ejs
```diff
@@ -0,0 +1,22 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <% include _head %>
+  </head>
+  <body class="homepage">
```
Manvel: We don't seem to use `homepage` class, guess we can get rid of it?

theme/less/homepage.less
```diff
@@ -0,0 +1,89 @@
...
+
+.box
+{
+  position: relative;
+  background-color: #f2f2f2;
+  border-radius: 8px;
+  border: 1px solid #dcdcdc;
+  padding: 10px 10px 60px 10px;
+  margin: 20px 0 0 0;
+  z-index: -1;
```
Manvel: This causes the links to be unclickable. Let's not use negative `z-index`.

Manvel: You can instead use positive `z-index` on the absolute `ul` element in navigation.

theme/less/_homepage.less
Manvel: Guess this one is still relevant?

.gitignore
naarrek: So, how to make it not to generate?

theme/less/_homepage.less
naarrek: No, I'm I think using `column-gap` on `main article`. This property is removed.

.gitignore
Manvel: I mean, you can remove this line, because `_content.css ` is being generated as part of `main.css` and `homepage.css`, as we don't explicitly include `content.css ` into the page, we don't need this exception in `.gitignore`, so if I understand correctly you can just remove this line.

naarrek: Although it's a partial file somehow it's generating in the public/css/ directory. I don't know how to deal with it.

Manvel: That's weird, I can't reproduce it on my side, can you confirm that after deleting the generated `public/css/content.css ` it's keep getting generated on build?

naarrek: Sorry for inconsistency, but I think deleting the file one more time resolved this issue.

public/css/homepage.css
naarrek: `git rm --cached public/css/homepage.css` resolved this I guess.

pages/index.html
```diff
@@ -0,0 +1,59 @@
+---
+layout: homepage
+title: CBA is an extension for chrome browser automation
+menu: Homepage
+---
+
+<img src="/images/Manvel_440x280_border.jpg" alt="CBA banner" />
```
Manvel: Detail: let's use more meaning-full image name here. `hero.jpg`, `cba-banner.jpg`.. Guess you can find something more meaningful.

pages/index.html
```diff
@@ -0,0 +1,59 @@
...
+    href="https://chrome.google.com/webstore/detail/jmbmjnojfkcohdpkpjmeeijckfbebbon"
+    target="_blank"
+    >CBA in chrome web store</a
+  >
+</section>
+
+<section class="box">
+  <h3>Documentation</h3>
+  <p>
+    <img src="/images/doc_grey50.png" />
```
Manvel: Detail: Let's not use too specific image name, i.e. "grey" and "50". Thinking:
- `/images/icon/document.png`
  - in this case we need to move all icons to specific folder
- `/images/document.png`
  - requiers fewer changes

Manvel (approved):
Good job @naarrek jan, I have added 2 last comments, about changing the filenames, feel free to address those and merge the PR afterwards.

Issue #10, opened 2020-04-09
Add content styles


Manvel:
# Background
Currently content styles are missing. Let’s first migrate those content styles from https://chrome-automation.com non index page that make sense and tackle styles for the index page and actual index page separately.

# What to change
Add content styles.

Manvel:
What to do with navigation inside of main:
- Suggestion is to wrap main content in `<article>`.
  - Currently on favor of this approach ^
- Alternative move `<nav>` item outside of main.

Issue #30, opened 2020-07-03
Introduce new front-matter properties for homepage content generation


naarrek:
### Background

We have already implemented styles for homepage. From content development point of view it will be more consistent to add several front-matter properties to make further content changes easier for homepage. 

### What to change

We can add properties for example like this:
```
index.md
---
title: Homepage
menu: Homepage
layout: homepage
hero-image: /images/Manvel_440x280_border.jpg
box-download-title: Download
box-download-content: Get chromium browser automation from google web store.
box-download-button: CBA in chrome webstore
...
---

1. CBA gives ability to automate chrome browser.
1. Ability to setup step-by-step code injections from Extensions popup interface.
1. Ability to Record interactions in the chrome browser and playing them.
...
```

PR #31, opened 2020-07-06
Issue #10 - Content Styles

naarrek:


theme/less/_content.less
```diff
@@ -2,3 +2,31 @@ main article img
...
 }
+
+article
+{
+  color: #444444;
+}
+
+article p, article ul, article ol
+{
+  font-size: 14px;
```
Manvel: Suggestion: Let's use REMs or EMs for font sizing, that way if user changes the font settings in the browser, browser will be able to scale the font. Otherwise fixed font size will be used. Same applies to other font sizes in current file.

pages/documentation/actions/change.md
```diff
@@ -8,13 +8,13 @@ Change action mostly are used on textboxes and textareas, they are Triggering fo
 
 Change action uses Jquery selector to Change the value of the DOM Element. Also users can use CBA placeholders:
 
-## Placeholders
+# Placeholders
```
Manvel: We already have `<h1>` set as title of the page, this suppose to be subtitle, so I'm not sure if this change is necessary, same applies to other pages where we might have `<h1>` used as subtitle.

pages/documentation/actions/bg-inject.md
```diff
@@ -14,7 +14,7 @@ Tips:
 - Stop project workflow when using asynchronous functions using sendBgInstruction variable "sendBgInstruction = false;"
 - Don't forget to continue the project workflow after stoping it `sendInstruction();`
 - use clipboard object for message passing `clipboard["myKey"] = value`
-- Jump to another action "actionToPlay(num);"
+- Jump to another action `actionToPlay(num);`
```
Manvel, line 17: Good job fixing this.

pages/documentation/actions/change.md
```diff
@@ -8,13 +8,13 @@ Change action mostly are used on textboxes and textareas, they are Triggering fo
...
 
-`<$unique=>` - Unique placeholder is generating random number (13 characters length maximum). So in the pic. example the action is used to find Dom Element with the ID of #sometextBox trigger focus event on it change the value to hello concatenated with 3 charecter length random number. 
+*`<$unique=>`* - Unique placeholder is generating random number (13 characters length maximum). So in the pic. example the action is used to find Dom Element with the ID of #sometextBox trigger focus event on it change the value to hello concatenated with 3 charecter length random number.
```
Manvel: Note: I don't have strong opinion about current change, but if you feel like we need to ephasize those items from senatic point of view, sounds good to me, just please don't use `em` for styling items(e.i. as italic).

theme/less/_content.less
```diff
@@ -2,3 +2,31 @@ main article img
...
+  margin-bottom: 14px;
+}
+
+article h2
+{
+  font-size: 28px;
+  font-weight: normal;
+  line-height: 130%;
+  margin-bottom: 10px;
+}
```
Manvel: I think there are still some some styles are missing:
- Let's add `h3` and `h4` at least, to provide more flexibility when creating pages.
- `<a>` tag has different colors
- `<pre>` Pre should be used when we are posting a code snippet, so styling it here would be great, in future we might also want to use [highligh.js](https://highlightjs.org/) for it.
- `<code>` thinking adding a light grey background might make content look like a snippet ?(in case you want to experiment)
- `` - I think it's bit tricky, we are using sometimes inside of `<p>`, sometimes inside it's own `<p>` tag, which currently shows them inconsistently. Wonder if we might make some content writing rules for that?

theme/less/_content.less
```diff
@@ -2,3 +2,31 @@ main article img
 {
   max-width: 100%;
 }
+
+article
+{
+  color: #444444;
```
Manvel: Suggestion: Let's specify default font-size here, as font-size is [inherited](https://developer.mozilla.org/en-US/docs/Web/CSS/inheritance) property it will be applied by default to all it's children.

theme/less/_content.less
```diff
@@ -2,3 +2,31 @@ main article img
...
+{
+  font-size: 14px;
+  margin: 0 0 20px 0;
+  line-height: 1.5;
+}
+
+article h1
+{
+  font-size: 35px;
+  font-weight: normal;
```
Manvel: Suggestion: Let's use numerical representation(e.i. `400`) of the `font-weight`, it's more handy when using external fonts, so you know exacly the weight of the fonts you need to import.

theme/less/_content.less
```diff
@@ -2,3 +2,31 @@ main article img
 {
   max-width: 100%;
 }
+
+article
+{
+  color: #444444;
+}
+
+article p, article ul, article ol
```
Manvel: In the Drupal version of the website `ul` and `ol` have top and bottom margins, I wonder if we need those here as well?

theme/less/_content.less
```diff
@@ -2,3 +2,31 @@ main article img
...
+{
+  font-size: 14px;
+  margin: 0 0 20px 0;
```
Manvel: Detail: Wonder if we want 20px bottom margin on the last element? that might provide unexpected results in general, I prefer using top margin if top most item in content (i.e. h1) has big margin bottom, in that case the margins might get merged, but not sure if it's same in our case.

This might not be a big deal though.. Leaving as "detail".

Manvel (changes requested):
@naarrek Thanks for the PR.

I have added comments, some of them we have already discussed in the chat, the others might be new to you.

As mentioned in the chat there are some inconsitencies with the Drupal version of the website, some inconsistencies seem to be intentational, the other might be not. Guess after the comments will be addressed things might be more clear for us.

pages/documentation/actions/change.md
naarrek: if you find those two pieces of code not very important to emphasize, I suggest not to do such way and to leave the code look like code and nothing more special. Or I should find alternative to `<em>`? For example, the `<i>` tag would be best choice i think considering this -> https://developer.mozilla.org/en-US/docs/Web/HTML/Element/em .

Manvel: Sorry, didn't mean to suggest using `<i>` instead, for emphasis we should use `<em>`, I meant to say that please don't use `<em>` only for styling, like making text "italic", but rather when you think we should emphasis specific part of content.

theme/less/_content.less
naarrek: We should use `em`s for `p, h1, h2, h3 ...` in that case to control font-sizes.

theme/less/_content.less
naarrek: You're right the top margins of  `p`  are collapsing  with bigger margin above. Also default margin of `p` is 1rem which is in most cases equal to 16px (by default browser settings). So this if this property will be removed nothing valuable will be changed.

theme/less/_content.less
naarrek: What you think is it  OK for these elements to have same font sizes and line-heights?

theme/less/_content.less
naarrek: Concerning to `` i think it's up to content writer to have it in `p` or separate. 
Or we can use same properties to display `` separately as with `p`.

theme/less/_content.less
Manvel: Right, then forget my previous comment. Also thinking if you use EMs, then we might want to use REM on current item, but if we use REMs everywhere we might not need that, so we know that the font-sizes are consistent with HTML element font-size and if a fixed font-size is specified on any parent of current element, content stays flexible and more accessible.

theme/less/_content.less
Manvel: For line-height to have same height as font-size `line-height: 1` can be used AFAIK. In general that's a UX question, so I would say rely on your own judgement, for some fonts having `line-height: 1` specified can make lines very close to each other affecting readability.

theme/less/_content.less
Manvel: Regarding `<pre>` We might want to create a creative solution in the future, thinking about having CBA UI in the background, so users know where to paste code, so I guess we can scratch using `highlight.js` in current iteration, maybe again some basic stylings can help.
```
Similar to this?
```

Manvel: > Concerning to  i think it's up to content writer to have it in p or separate

Right. Maybe make sense to check how other websites are dealing with this, it's still a tricky question.

naarrek: what kind of style should we apply on `<pre>`?

theme/less/_content.less
naarrek: I think we should use `rem`s for `article` and `em`s for it's children, this way the browser will be able to scale the the `article` font sizes and it's children indirectly.

theme/less/_content.less
naarrek: > 
> Right. Maybe make sense to check how other websites are dealing with this, it's still a tricky question.

In this post google says don't put the `` inside `<p>` https://developers.google.com/style/images
But as I looked in websites like google or VSCode supports I found images in `<p>`s
Like here:
https://support.google.com/webmasters/answer/114016?hl=en
And here:
https://code.visualstudio.com/docs/editor/userdefinedsnippets

theme/less/_content.less
Manvel: > Also default margin of p is 1rem

Is it? I wouldn't rely on it, usually it's up for the browser vendors to specify those. 

I found so far that users are sometimes refering to [this resource](https://www.w3.org/TR/CSS21/sample.html), but as the page says is "informative, not normative".

theme/less/_content.less
```diff
@@ -1,4 +1,62 @@
...
+
+article p:first-of-type img
+{
+  margin-top: 1em;
+}
+
+article h1
+{
+  font-size: 2.19em;
+  line-height: 140%;
```
Manvel: Detail: Here you are using percentage as value, but above numeric multiplier for `line-height`. Let's use one or another for consistency. I more lean towards numeric multiplier, as those more similar to EMs and REMs multiplication logic.

theme/less/_content.less
```diff
@@ -1,4 +1,62 @@
...
 {
   max-width: 100%;
 }
+
+article p:first-of-type img
+{
+  margin-top: 1em;
+}
+
+article h1
```
Manvel, line 29: My system doesn't have "PT Sans" installed and fonts are rendered differently on for me in comparison to the Drupal version of the website, that's seem to be because in Drupal version Fonts are being requested from [https://fonts.google.com/](https://chrome-automation.com/themes/skeletontheme/css/style.css).

We can either fix font issue in current issue or separately. To fix it, I think first we need to ask the questions:
- Do we want to use PT Sans?
- How we want to import them?

## Using PT Sans

Using PT Sans sounds like a low hanging fruit and we can definately use it. We should for future take into consideration that it's only supported in `cyrillic` and `latin`. In general fonts and multi-language support is a quite complex topic, I have briefly touch that in [current talk - from 10:45min](https://youtu.be/YpRc79o-QHM?t=645).

Usually when I decide about fonts I open https://fonts.google.com/ and sort them by popularity, Roboto, Source Sans Pro and Open Sans, are those I was working lately, it's a matter of choice and preference, thinking it's best to find one matching closest to PT Sans, or just go with PT Sans. and call it a day.

## Importing fonts

We should ensure that the font we are using is woff2 format, as it's currently most advanced compression for those. Also it's always a good idea to host fonts on your own server rather than on Google's, that also might help with the privacy concenrs, because if fonts are downloaded from https://fonts.google.com, then we basically sending request to another server. In most of the cases it's not a problem, but regarding the privacy I think if you can avoid then you should.

Here are couple of examples of fonts implementation with CMintS:
- [Multilanguage](https://github.com/cmints/website/blob/master/theme/less/_fonts.less)
- [Single language](https://github.com/Manvel/blog/blob/master/theme/less/fonts.less)

For fonts conversion I guess bet tool would be [font-tools](https://github.com/fonttools/fonttools), but usually you can get ready woff2 fonts from https://fonts.google.com, I guess in the video above there are also solutions for that.

Manvel (commented):
@naarrek Jan thanks for the PR.

I have made only 1 detail comment and some guides for fonts implementation, if you decide implement fonts separately I'm happy to approve this PR (ideally if/when detail comment is also addressed).

theme/less/_content.less
```diff
@@ -1,32 +1,62 @@
...
+}
+
+article h4
+{
+  font-size: 1.12em;
+  line-height: 110%;
+}
+
+article a:hover, article a:visited:hover
+{
```
naarrek, line 70: StyleLint test is not passing because of this part. If we putting this part we're getting this error:
`54:1  ×  Expected selector "article a:focus" to come before selector "article a:visited:hover"   no-descending-specificity`
And getting similar error when I put this below it:
`53:1  ×  Expected selector "article a:hover" to come before selector "article a:visited:focus"   no-descending-specificity`

theme/less/_body-wrapper.less
```diff
@@ -12,6 +12,30 @@ body > div
   padding: 20px 30px;
 }
 
+@font-face 
+{
+  font-family: "Source Sans Pro";
+  font-weight: 400;
+  font-style: normal;
+  src: local("Source Sans Pro Regular"), local("SourceSansPro-Regular"), url("/fonts/SourceSansPro-Regular.woff2") format("woff2");
```
Manvel: Suggestion: Please name the font according to it's height, so we know what we are loading, ex.: `fonts/source-sans-pro/400.woff2` in you are uploading font with that weight.

Manvel: You can also use `latin` in the path/filename, or make a comment in `@font-face`, so we know what we have here.

theme/less/_body-wrapper.less
```diff
@@ -12,6 +12,30 @@ body > div
...
+}
+
+@font-face 
+{
+  font-family: "Source Sans Pro";
+  font-weight: 700;
+  font-style: bold;
+  src: local("Source Sans Pro Bold"), local("SourceSansPro-Bold"), url("/fonts/SourceSansPro-Bold.woff2") format("woff2");
```
Manvel: I get current error:
```
Failed to decode downloaded font: http://127.0.0.1:3000/fonts/SourceSansPro-Bold.woff2
```

And font isn't loaded, only this font have that issue though, both in Firefox and Chrome.

theme/less/_body-wrapper.less
```diff
@@ -12,6 +12,30 @@ body > div
...
+  font-weight: 400;
+  font-style: normal;
+  src: local("Source Sans Pro Regular"), local("SourceSansPro-Regular"), url("/fonts/SourceSansPro-Regular.woff2") format("woff2");
+}
+
+@font-face 
+{
+  font-family: "Source Sans Pro";
+  font-weight: 700;
+  font-style: bold;
```
Manvel: Not really sure if we need to specify that the font-size is bold, in general thinking we can ignore `font-style ` rule and use `font-weight: 700;` to resolve the one we need.

theme/less/_body-wrapper.less
```diff
@@ -1,6 +1,6 @@
 body
 {
-  font-family: "PT Sans", "HelveticaNeue", "Helvetica Neue", Helvetica, Arial, sans-serif;
+  font-family: "Source Sans Pro";
```
Manvel: Let's still specify fallback fonts, because if there is a network issue we might be able to use a local font, as it's a font from serif family, keeping `sans-serif ` as fallback or `Helvetica, Arial, sans-serif ` as it was before should be good to go. So the later one means, if `Source Sans Pro ` can't resolved for the character use  Helvetica, if no Helvetica on the system use Arial, if no Arial, then resolve to a sans-serif font.

theme/less/_body-wrapper.less
```diff
@@ -12,6 +12,30 @@ body > div
...
+  font-weight: 700;
+  font-style: bold;
+  src: local("Source Sans Pro Bold"), local("SourceSansPro-Bold"), url("/fonts/SourceSansPro-Bold.woff2") format("woff2");
+}
+
+@font-face 
+{
+  font-family: "Source Sans Pro";
+  font-weight: 400;
+  font-style: italic;
```
Manvel: Wonder if nowdays browsers are smart enough that can immitate `italic` without downloading specific font for that, make sense to try.

Manvel (changes requested):

theme/less/_content.less
Manvel: Guess we can overwrite the `no-descending-specificity` rule in stylelint.

theme/less/_body-wrapper.less
naarrek: I think this works well.

theme/less/_content.less
naarrek: There is similar issue reported here -> https://github.com/stylelint/stylelint/issues/3052 and here -> https://github.com/stylelint/stylelint/issues/2489. They are explaining that this isn't a bug, but I still not sure how we can resolve this issue.

theme/less/_body-wrapper.less
Manvel: Suggestion: let's not use uppercase in the file names, where possible. That way we will be consistent with all other files we have.

theme/less/_content.less
```diff
@@ -1,4 +1,62 @@
...
+  margin-top: 1em;
+}
+
+article h1
+{
+  font-size: 2.19em;
+  line-height: 1.4em;
+}
+
```
Manvel, line 45: Meeting-note:
- Let's specify margin for `article p` where margin-bottom is smaller, for the cases when we have following ul.
- And also specify margins for article headings.

theme/less/_content.less
```diff
@@ -1,4 +1,62 @@
...
+  color:#000000;
+  text-decoration: none;
+}
+
+article a:focus, article a:visited:focus
+{
+  color:#000000;
+}
+
+article code
```
Manvel, line 81: Meeting-notes: 
- Let's add small padding
- Let's use same background on pre

theme/less/_content.less
```diff
@@ -1,4 +1,62 @@
-main article img
+article
```
Manvel, line 1: meeting-notes:
- Bold fonts seem to be too bold in comparison to Drupal, we can try using 600 semi-bold instead as a compromise.
- And let's try move content up, if possible.

theme/less/_content.less
```diff
@@ -16,47 +16,72 @@ article p:first-of-type img
...
+article h3 + ul
+{
+  margin-top: 0.5em;
+}
+
 article h1
 {
   font-size: 2.19em;
   line-height: 1.4em;
+  margin: 0.3em 0 0 0;
```
naarrek: 0.3em here can be for example increased. I did this to be symmetric with sidebar but it's not necessary change.

theme/less/_header.less
```diff
@@ -8,15 +8,18 @@ header > p
   color: #7d7d7d;
   font-size: 24px;
   letter-spacing: -1px;
+  line-height: 25px;
+  margin-bottom: 0;
```
naarrek: Here `line-height` controls distance between title and subtitle of the site. Also distance between subtitle and navigation I left to control navigation `ul` margin considering it "more static (not changeable)" than subtitle.

theme/less/_content.less
```diff
@@ -16,47 +16,72 @@ article p:first-of-type img
   margin-top: 1em;
 }
 
+article p
+{
+  margin: 0.5em 0 0 0;
+}
+
+article p + ul
```
Manvel: I would avoid using "+"  selector here, you are explicitly making an exception for p and ul, that might lead to having too many of those exceptions. Like the one below, what about `h4` for example?

theme/less/_body-wrapper.less
```diff
@@ -22,8 +22,8 @@ body > div
 @font-face
```
Manvel, line 22: Seems like you have added `public/fonts/SourceSans-pro-latin/400-Italic.woff2` back again, are we using it again?

theme/less/homepage.less
```diff
@@ -85,4 +94,20 @@ main article > *:first-child
   {
     grid-column: 1;
   }
+  
+  .box:nth-of-type(2)
```
Manvel: Maybe addressing the issue below we will avoid being so specific, by adding a specific style to the second box, as we would like to have consistency and when another block is added we don't care about it's number, I think we can avoid this.

theme/less/homepage.less
```diff
@@ -85,4 +94,20 @@ main article > *:first-child
...
+  {
+    margin-top: 15px;
+  }
+
+  main article ol
```
Manvel: That's another example why content element seem to be off for layouting. Maybe we can add `.cell` class, or `main article > * {margin-top:15px}`, to make it clear that the cells of your grid layout table suppose to have 15px distance from each other, to avoid specifying `margin-top: 15px`, for each item individually. Unless the distances are different, also I'm not sure why `15px` seems like in the Drupal version the distance is different?

Manvel (commented):
Narek Jan, I have added couple of suggestions here -> https://github.com/browser-automation/cba-website/pull/32.

I think we have discussed during the last meeting that we can try making the implementation more closer to the Drupal page, if you might have a chance, can you please have a look and let me know if you agree with the decisions, you can find list of adjustments in the PR description.

theme/less/_content.less
Manvel: Fixed in -> https://github.com/browser-automation/cba-website/pull/32

theme/less/_body-wrapper.less
naarrek: I think I didn't remove it ever.

theme/less/_content.less
naarrek: This I think is done.

theme/less/_body-wrapper.less
Manvel: I thought [we agreed](https://github.com/browser-automation/cba-website/pull/31#discussion_r453766244) to remove italic fonts, might be that I miss-interpreted your comment above.

So if we don't need them, please remove italic fonts, so pages won't need to download more data.

Manvel: I have noticed that you removed the italic font, but now there are still 4 fonts, 2 bold and 2 normal weight.

naarrek: It's a windows bug after renaming `SourceSans-Pro-latin` to `SourceSans-pro-latin`:
```
public/fonts/SourceSans-Pro-latin/400.woff2
public/fonts/SourceSans-Pro-latin/600-bold.woff2
public/fonts/SourceSans-pro-latin/400.woff2
public/fonts/SourceSans-pro-latin/600-bold.woff2
```

naarrek: I've addressed this issue in the last [commit](https://github.com/browser-automation/cba-website/pull/31/commits/18512409c7be0dd7546e96036fbdeb7b273afb42).

theme/less/_content.less
Manvel: I think the way it is right now is fine.

theme/less/homepage.less
```diff
@@ -32,7 +52,6 @@ main article > *:first-child
   border-radius: 8px;
   border: 1px solid #dcdcdc;
   padding: 10px 10px 60px 10px;
-  margin: 20px 0 0 0;
 }
 
 .box img
```
Manvel, line 58: - Icon margin-top needs to be set to 0
  - Use before
  - add class to icon
  - use important
- Buttons padding and fonts are different from Drupal

theme/less/homepage.less
```diff
@@ -20,6 +20,26 @@ main article
   column-gap: 20px;
```
Manvel, line 20: header h1:
- Is smaller than in drupal
- Missing line-height to adjust subheader

theme/less/_navigation.less
```diff
@@ -12,6 +12,7 @@
   border: #F5F5F5 solid  1px;
```
Manvel, line 12: - `.current-menu`(selected) item is white in Drupal.
- arrow icon needs to be adjusted to middle.
- Header menu paddings are different.
- Sub menu items font is smaller on Drupal.
- "Homepage" is "Home" on Drupal

theme/less/homepage.less
```diff
@@ -20,6 +20,26 @@ main article
...
 }
 
+main article ol
+{
+  line-height: 1.5rem;
```
Manvel: Suggestion:
```
main article ol
{
  line-height: 1.4rem;
}

article
{
  font-size: 0.85rem;
}
```

Manvel: If you like content being 0.9rem more on other pages apply font-size change only on homepage.

theme/less/_header.less
```diff
@@ -8,15 +8,17 @@ header > p
   color: #7d7d7d;
```
Manvel, line 8: Differences in mobile:
- line-height and font-sizes
  - heading
  - Sub heading
  - article heading (line-height)

theme/less/homepage.less
naarrek: In the case of `:before`  we need to specify particular box in styles for each image via let's say `:nth-child()` selector which is not the best solution considering that count of boxes can be changed in future. Or we can give classes to boxes and use images with `:before`. Which is optimal solution in case or boxes as they are complex objects and having a class for each one i think is OK.

naarrek: In Drupal site the font-weight is 900, do I need to add one more font file, or using here 700 (for consistency sake) is OK , what you think?

theme/less/homepage.less
naarrek: Done.

theme/less/homepage.less
Manvel: > classes to boxes and use images with :before

I think I have classes in mind when I was proposing, something along the lines `.icon .download`, `.icon-download`, i.e.: `<p class="icon download">`.

You can also put boxes outside of the `<article>`, that way content styles won't be applied to them and you will have all the freedom designing boxes.
```
<article>
  <h2></h2> 
  <img>
  <ol></ol>
</article>
<section class="box">
  ...
</section>
...
```

Manvel: > In Drupal site the font-weight is 900, do I need to add one more font file, or using here 700 (for consistency sake) is OK , what you think?

I don't have strong opinion, but I feel like Source Sans Pro's `400` is more closer to the Drupal's "normal" font, which is used on header, rather than Source Sans Pro's `600`, which is too bold in comparison. I can't say for sure if Source Sans Pro's `500` will fit here better, but that's only mediocre solution we can have with Source Sans Pro, though I'm not sure if that worth it as we will download whole font to show one sentence.

On a relevant note I did notice that Fonts you added are to divergent in sizes, which is bit suspicious - 600 15kb, 400 87kb, make sense to double check if 400 contains on latin.

Following up: If it's about adding another 15kb font, then maybe it worth experiment with 500.

naarrek: The problem with this solution is that `<article>` is in template `.ejs` file. Maybe we should go with simplest option and use here `!important`?

Manvel: > The problem with this solution is that <article> is in template .ejs file

Right, I think it would be less problem after we address -> https://github.com/browser-automation/cba-website/issues/30 or you can add box properties right away and move `.box`-es outside of article as part of current PR.

> Maybe we should go with simplest option and use here !important

Fine by me.

theme/less/homepage.less
naarrek: Can you check please this values one more time? This is the result ->
image

theme/less/homepage.less
naarrek: > 
> Following up: If it's about adding another 15kb font, then maybe it worth experiment with 500.

It turned out that the real size of latin 400 font is again 15kb, but the version of 500 does not exist in google fonts repository. And consequently the font generator does give nothing -> https://fonts.googleapis.com/css?family=source+sans+pro:500&subset=latin

theme/less/_navigation.less
```diff
@@ -3,6 +3,7 @@
 #main-navigation 
 {
   background: linear-gradient(#fcfcfc, #ececec);
+  border-color: #EAEAEA #F5F5F5 #F5F5F5 #F5F5F5;
```
naarrek: this is not necessary,  I'll remove it.

theme/less/_navigation.less
```diff
@@ -93,13 +95,16 @@
...
+  vertical-align: sub;
 }
 
-#main-navigation ul ul li.has-children > a:after
+#main-navigation ul ul li.has-children a:after
 {
   position: absolute;
   right: 10px;
+  top: 33%;
+  padding-right: 10px;
```
naarrek: Maybe it's right to do here `padding: 0 10px 0 0`.

theme/less/_navigation.less
```diff
@@ -93,13 +95,16 @@
 #main-navigation li.has-children > a:after
 {
   content: url('/images/arrow_bullet.png');
-  padding-left: 10px;
+  padding-left: 11px;
```
naarrek: Maybe it's right to do here `padding: 0 0 0 11px`.

theme/less/_navigation.less
```diff
@@ -93,13 +95,16 @@
...
+  vertical-align: sub;
 }
 
-#main-navigation ul ul li.has-children > a:after
+#main-navigation ul ul li.has-children a:after
 {
   position: absolute;
   right: 10px;
+  top: 33%;
```
naarrek: `vertical-align:middle` or `vertical-align:sub` doesn't work on #main-navigation ul ul li.has-children a:after because it's positioned absolutely. So I added there top with percentage units.

theme/less/_navigation.less
```diff
@@ -93,13 +95,16 @@
 #main-navigation li.has-children > a:after
 {
   content: url('/images/arrow_bullet.png');
-  padding-left: 10px;
+  padding-left: 11px;
+  vertical-align: sub;
```
naarrek: This aligns `a:after` as subscript of the text. Instead of this could be for example  `vertical-align` with negative `em` or percentage values to align image properly.

theme/less/_navigation.less
```diff
@@ -93,21 +104,23 @@
...
```
Manvel, line 108: This is how it looks like, on my side.

Screen Shot 2020-08-04 at 10 15 29 PM

switching this to `vertical-align: middle;` seem to fix the arrow to be aligned in the middle of the text.

theme/less/_navigation.less
Manvel: This is how it looks like, on my side.

Screen Shot 2020-08-04 at 10 15 29 PM

switching this to `vertical-align: middle;` seem to fix the arrow to be aligned in the middle of the text.

theme/less/_navigation.less
Manvel: It will have same effect. Having `padding-left` seem to be fine.

theme/less/_navigation.less
Manvel: Right, what about using:
```
display: flex;
justify-content: space-between;
```

On `<a>` element instead?

theme/less/_navigation.less
Manvel: Same as here https://github.com/browser-automation/cba-website/pull/31#discussion_r465275029

theme/less/homepage.less
Manvel: I see, seems like Source Sans Pro doesn't have 500 weight

theme/less/homepage.less
Manvel: Seems like you are still using `0.9rem` instead of `0.85rem`, but I thinking keeping `0.9rem` is fine, now after seeing all other parts being more aligned with the Drupal version.

theme/less/_navigation.less
Manvel: Also seems like items without side arrow are missing some paddings.

theme/less/_navigation.less
naarrek: It seems `display: inline-block` on `#main-navigation ul ul a` solves this vertical height problem. No need to change line-height for `<a>` , I guess...

theme/less/homepage.less
naarrek: True. But 0.92rem gives result similar to Drupal version.

theme/less/_navigation.less
Manvel: What about the absolute positioning? Not sure how solution with `display:  inline-block ` can solve it, might be I didn't understand your comment. But feel free to apply your proposed changes and we can revisit this.

theme/less/homepage.less
Manvel: Calculated line height in Drupal website is `21px`, in CMintS version is `22.4px`, Increasing line height can't give similar result, unless you are planing to adjust height by changing `line-heigh` which is not a good idea I guess.

theme/less/_navigation.less
```diff
@@ -104,16 +110,14 @@
...
 }
 
 #main-navigation ul ul li.has-children a:after
 {
-  position: absolute;
-  right: 10px;
-  top: 33%;
-  padding-right: 10px;
+  line-height: 22px;
+  padding: 0 13px 0 0;
```
Manvel: Suggestion:
```
  padding: 0 13px;
  height: 15px;
```

`line-height: 22px;` makes sub-items height bigger. What you want to do here is to make the height of the icon equal to the image height and align the item to the middle.

Let's make the paddings equal from both sides, it will prevent text comming to close to the icon and also in future if we decide to support RTL languages, might be a good idea to make it even.

theme/less/homepage.less
```diff
@@ -18,7 +18,7 @@ main article
   display: grid;
   grid-template-columns: 1fr 1fr;
   column-gap: 20px;
-  font-size: 0.9rem;
+  font-size: 0.92rem;
```
Manvel: see -> https://github.com/browser-automation/cba-website/pull/31#discussion_r466997197

theme/less/_navigation.less
Manvel: I see you still changed to your suggestion. Anyway it's not a big deal, though I feel more in favor of left padding, if there is no such thing as "padding-end"

theme/less/_navigation.less
Manvel: LGTM.

theme/less/_navigation.less
Manvel: LGTM.

Manvel (approved):
LGTM.

PR #32, opened 2020-07-23
Couple of implementation suggestions from Manvel

Manvel:
Current PR contains:
- Adjustements for Drupal website similarity
  - Made h1 slimmer.
  - Increased p and ul margin-top distance
  - Decreased content font-size
  - added img width in content
- Added paddings to inline code block.
- Content markup adjustements
  - Missing p tags
  - heading inheritance adjustments

### DIFF
--- a/pages/documentation/play.md
+++ b/pages/documentation/play.md
@@ -6,4 +6,6 @@ weight: 5
 
 To start automation user basically need to select the project and click "play" button. After clicking "play" button actions begin to be injected to content script one by one, and you can follow the workflow process in actions datagrid, also you can see "play" message appear under the CBA icon that mean that Chromium Browser Automation is in the process of playing your scenarios, So in case if the workflow will stuck somewhere you always can click on "stop" button. "stop" Button stops the project workflow. Also user can mention how many times he want project to be played by changing the value in the textbox next to "play" button. 
 
-![](/images/play-times.jpg)
+<p>
+  <img src="/images/play-times.jpg" width="180px">
+</p>
--- a/pages/documentation/recording.md
+++ b/pages/documentation/recording.md
@@ -6,4 +6,6 @@ weight: 4
 
 CBA store all information in browser Local Storage so that why after closing the chrome browser and restarting the computer records are still available. Also you can record your interaction by clicking the "rec" button. After clicking record button "rec" message appear under CBA icon that mean that you can start to interact with web page. After finishing the interaction just click on "stop" button. Now you can manage and edit your automation. 
 
-![](/images/recording.jpeg)
+<p>
+  <img src="/images/recording.jpeg" width="200">
+</p>
--- a/theme/less/_content.less
+++ b/theme/less/_content.less
@@ -1,6 +1,6 @@
 article
 {
-  font-size: 1rem;
+  font-size: 0.9rem;
   color: #444444;
   font-weight: 400;
   line-height: 1.5;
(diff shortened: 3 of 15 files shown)


theme/less/_content.less
```diff
@@ -18,45 +18,47 @@ article p:first-of-type img
 
 article p
 {
-  margin: 0.5em 0 0 0;
+  margin: 14px 0 0 0;
```
Manvel, line 21: TODO: Use EMs instead for consistency.

pages/documentation/play.md
```diff
@@ -6,4 +6,6 @@ weight: 5
 
 To start automation user basically need to select the project and click "play" button. After clicking "play" button actions begin to be injected to content script one by one, and you can follow the workflow process in actions datagrid, also you can see "play" message appear under the CBA icon that mean that Chromium Browser Automation is in the process of playing your scenarios, So in case if the workflow will stuck somewhere you always can click on "stop" button. "stop" Button stops the project workflow. Also user can mention how many times he want project to be played by changing the value in the textbox next to "play" button.
```
naarrek, line 8: Here we have unnecessary paragraph I guess.

pages/documentation/recording.md
```diff
@@ -6,4 +6,6 @@ weight: 4
 
 CBA store all information in browser Local Storage so that why after closing the chrome browser and restarting the computer records are still available. Also you can record your interaction by clicking the "rec" button. After clicking record button "rec" message appear under CBA icon that mean that you can start to interact with web page. After finishing the interaction just click on "stop" button. Now you can manage and edit your automation.
```
naarrek, line 8: Here also we have unnecessary paragraph I guess.

theme/less/_content.less
```diff
@@ -18,45 +18,47 @@ article p:first-of-type img
...
 }
 
-article p + ul
+article ul
 {
-  margin-top: 0.3em;
+  margin: 20px 0 0 0;
```
naarrek, line 26: Isn't 20px too much? according to our discussion we need to decrease this value.

theme/less/_content.less
```diff
@@ -18,45 +18,47 @@ article p:first-of-type img
...
 
-article h3 + ul
+article h1
 {
-  margin-top: 0.5em;
+  font-size: 2.4em;
+  line-height: 1.4em;
+  margin: 20px 0 0 0;
+  font-weight: 300;
+  color:#181818;
```
naarrek, line 35: We have this color also on `h2, h3, h4, h5, h6` on Drupal site.

theme/less/homepage.less
```diff
@@ -23,10 +23,16 @@ main article
...
+  align-self: center;
 }
+
 main article h1
 {
-  margin: 6px 0 6px 0;
+  margin: 15px 0 6px 0;
+  font-weight: 300;
+  font-size: 35px;
+  color: black;
```
naarrek, line 35: I think we can here define black color as `color: #181818` for consistency.

theme/less/_content.less
Manvel: Yes, previously the content text was pushing the footer, now after decreasing the footer size we don't have that issue. AFAIK that was the case we decided to decrease. Currently the size is seem to be consistent with the size `<p>` is pushing below `<ul>` elements in the content in Drupal.

theme/less/_content.less
Manvel: I have left it more grey-ish on other heading elements, because in the Drupal the font-weight is lower and having a bit of grey-ish sub headers, felt to me less distracting when comparing the Drupal and current website, it feels more "calm", as a compromise to bold weight.

theme/less/homepage.less
Manvel: Good catch, agree.

pages/documentation/play.md
Manvel: AFAIK, when you have a block level element proceeding the new line, then an empty wrapper won't be generated, unless you add a content there.

theme/less/_content.less
naarrek: It's not only consistency problem. Do we want margins, paddings and line-heights to be scaled with `<html>` font-size? If yes, we can use here and in other places EMs. What you think?

Manvel: That's a good question, usually I used Pixel for margins and paddings, but EMs or REMs on font-sizes for simplicity in calculation on other projects. I think in general It make sense that the margins are also adjusted accordingly, but sometimes it might make sense to keep in simple and use PX to avoid unexpected behavior with side margins for example.

Short: Both will work, but let's make other margins and paddings following same pattern(either use PX or EMs) in `_context.less`.

naarrek: OK, to avoid unexpected layouts I'll use Pixels for margins and paddings. In other cases I'll use REM and EMs.

theme/less/_content.less
naarrek: I'll leave it like that to be consistent with Drupal version.

theme/less/_content.less
naarrek: Not really sure about this. Well, let me rely on your decision and intuition then...

Issue #38, opened 2020-09-12
Improve website metrics


Manvel:
# Steps to reproduce

1. Navigate to https://web.dev/measure/
1. Run audit for https://chrome-automation.netlify.app/

## Observed behavior

The website still have some weak measures:

<img width="1027" alt="cba-website-metric" src="https://user-images.githubusercontent.com/716969/92993721-f22e0d00-f504-11ea-9d62-a953be3f79e7.png">

## Expected behavior

The website meets all the requirements mentioned in the metric description.

#### Accessibility
- [x] - [Image elements do not have `[alt]` attributes](https://web.dev/image-alt/) - https://github.com/browser-automation/cba-website/pull/39
- [x] - [Form elements do not have associated labels](https://web.dev/label/) - https://github.com/browser-automation/cba-website/pull/43
- [x] - [Background and foreground colors do not have a sufficient contrast ratio.](https://web.dev/color-contrast/) - https://github.com/browser-automation/cba-website/pull/43
- [x] - [Heading elements are not in a sequentially-descending order](https://web.dev/heading-order/) - https://github.com/browser-automation/cba-website/pull/43
- [x] - [`<html>` element does not have a `[lang]` attribute](https://web.dev/html-has-lang/) - https://github.com/browser-automation/cba-website/pull/43
#### Performance
- [ ] - [Eliminate render-blocking resources](https://web.dev/render-blocking-resources/) - https://github.com/browser-automation/cba-website/pull/45
- [ ] - ~[Efficiently encode images](https://web.dev/uses-optimized-images/)~
  - (The desired optimization by lighthouse decreases the image quality noticeably).
- [ ] - ~[Serve images in next-gen formats](https://web.dev/uses-webp-images/)~
  - (Safari doesn't support next-gen formatting, having fallback for each of images, might be too much effort right now.)
####  Best Practices
- [x] - [Links to cross-origin destinations are unsafe](https://web.dev/external-anchors-use-rel-noopener/) - https://github.com/browser-automation/cba-website/pull/46
- [ ] - ~[Serves images with low resolution](https://web.dev/serve-responsive-images/)~
  - (Serving multiple images might provide too much maintenance effort on this stage)
#### SEO
- [ ] - [Document does not have a meta description](https://web.dev/meta-description/)
- [x] - [Image elements do not have `[alt]` attributes](https://web.dev/image-alt/) - https://github.com/browser-automation/cba-website/pull/39

PR #43, opened 2020-09-13
#38 - Fixed accessibility issues reported by lighthouse

Manvel:
- Added `aria-label` to page navigation selectbox on mobile.
- Used `<h2>` instead of `<h3>` in `.box` element on homepage.
- Adjusted subheading contrast to meet minimal requirements.
- Adde `lang="en"` to document.

### DIFF
--- a/theme/less/homepage.less
+++ b/theme/less/homepage.less
@@ -62,8 +62,10 @@ main article > *:first-child
   margin: 0!important;
 }
 
-.box h3 
+.box h2 
 {
+  font-size: 1.4em;
+  line-height: 120%;
   margin: 0;
 }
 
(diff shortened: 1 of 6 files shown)


theme/less/homepage.less
```diff
@@ -62,8 +62,10 @@ main article > *:first-child
   margin: 0!important;
 }
 
-.box h3 
+.box h2 
 {
+  font-size: 1.4em;
+  line-height: 120%;
```
Manvel, line 68: Note: I did notice that we are using weird `font-size` and `line-height` sequence, where line-height is using smaller em value, but produces higher result, probably because higher line-height is set on parent. Thinking switching to percentage, make more sense to avoid such issues. Wonder if we should address that everywhere else?

Manvel: Created issue for that -> https://github.com/browser-automation/cba-website/issues/44

naarrek (approved):

PR #47, opened 2020-09-20
No issue - Added privacy policy page

Manvel:


pages/privacy-policy.md
```diff
@@ -0,0 +1,36 @@
+---
+title: Privacy Policy
+permalink: privacy
+---
+
+## Analytics
+
+CBA is user centric product, CBA is only using Google Analytics to collecting
```
naarrek: CBA is using only Google Analytics

pages/privacy-policy.md
```diff
@@ -0,0 +1,36 @@
...
+minimal data only to understand how the product is used and find generic traffic
+pattern in chrome-automation.com website. No other data is collected and only
+used internally to understand how to improve the product. You can find summary
```
naarrek: used only internally

Manvel: As we would like to emphasize that it's only used internally and never shared with 3-rd parties. Guess using only in the beginning might be more explanatory in this case?

pages/privacy-policy.md
Manvel: Used `CBA is user centric product, CBA is using Google Analytics only to collect minimal data, just to understand how the.` instead, as agreed. As it's more informative and makes clearer, that we collect minimal data *only*.

PR #61, opened 2023-09-07
Rename index.md to index.md

leonardoadame:


pages/documentation/index.md
```diff
@@ -19,4 +19,4 @@ CBA is automation system for chrome browser which is designed to help people wit
 - Record
 - Edit automation
 - Manage
-- and Play
\ No newline at end of file
+- and Play
```
Manvel, line 22: @leonardoadame thanks for the contribution, think you added a new line, to the end of the current file, what about adding it to all `.md` files missing it and make it a rule? Currently think there is no consistency, which is, fine. But happy to accept when the contribution adds consistency to all relevant files missing that.

Manvel:
Closing because of inactivity, please re-open if this is relevant.

